PR #35, opened 2018-10-04
fixed building falcon client

tutok:
fixed building falcon client

### DIFF
--- a/packages/falcon-client/src/buildTools/razzle/falconClientPlugin.js
+++ b/packages/falcon-client/src/buildTools/razzle/falconClientPlugin.js
@@ -233,6 +233,7 @@ module.exports = appConfig => (config, { target, dev }, webpackObject) => {
   };
 
   setEntryToFalconClient(config, target);
+  // make sure that webpack handle @deity/falcon-client from shop directory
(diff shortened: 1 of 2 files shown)


packages/falcon-client/src/buildTools/razzle/falconClientPlugin.js
```diff
@@ -213,9 +214,27 @@ module.exports = appConfig => (config, { target, dev }, webpackObject) => {
   };
 
   setEntryToFalconClient(config, target);
-  extendBabelInclude([paths.falconClient.appSrc, path.join(paths.resolvePackageDir('@deity/falcon-ui'), 'src')])(
-    config
-  );
+  extendBabelInclude([
+    paths.falconClient.appSrc,
+    /\/@deity\/falcon-client\//,
+    path.join(paths.resolvePackageDir('@deity/falcon-ui'), 'src')
```
Borales: @tutok developers may not use falcon-ui package, will it throw an error in that case?

tutok: @Borales yes you are right, it is removed already ;)

PR #47, opened 2018-10-05
Improvements for README of create-falcon-app

mateuszwozniak:


Borales:
@mateuszwozniak please update your branch from master

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,69 @@
...
+and the client in the second terminal:
+```bash
+cd my-app/client
 npm start
 ```
+then open http://localhost:3000 in your browser
 
-or with the `yarn create` command:
+## Longer version
+You can use either npx (installed with npm) or yarn to generate the application.
```
Borales: @mateuszwozniak general question/comment - should we include all names of the tools with backquotes?

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,69 @@
...
-or with the `yarn create` command:
+## Longer version
+You can use either npx (installed with npm) or yarn to generate the application. 
 
+If you prefer npx use the following command:
+```bash
+npx create-falcon-app my-app
+```
+
+and if you prefer yarn use `yarn create` with the following syntax (notice space between `create` and `falcon-app`:
```
Borales: @mateuszwozniak maybe indent with `>` block?

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,69 @@
...
-You can also initialize a project using one of our [example](./../../examples) templates:
+That command will create folder `my-app` with 2 folders inside: 
+```
+my-app/
+ |- server
+ |- client
+```
+
+Folder `server` contains server which is responsible for communication with external APIs and services.
+Server `client` contains service responsible for front-end part of the application.
```
Borales: @mateuszwozniak "Server" -> "Folder"

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,69 @@
 # Create Falcon App
 
-Create your own Falcon-based application with a few commands:
+This CLI tool helps you to create your application based on [DEITY Falcon](https://github.com/deity-io/falcon) with just one command.
```
Borales: @mateuszwozniak should we add https://user-images.githubusercontent.com/1118933/44960665-c9311580-af03-11e8-9475-565cd0d85eda.gif to this file?

cane-lagerwaard: @Borales is this the latest version? Also it doesn't use ```npx```

mateuszwozniak: nope, becuase:
1. we don't want people to install create-falcon-app locally but use it via npx
2. it's hard to show whole process (generate the app and start both client and server). We don't want to confuse people with gif that doesn't show exactly what should be done

packages/create-falcon-app/README.md
mateuszwozniak: We could do that but it won't work for all the cases (e..g sometimes we might want to use link to the package). But generally I think it's a good practice, as the name stands out from the text  and it's easier to understand.

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,74 @@
...
 
-You can also initialize a project using one of our [example](./../../examples) templates:
+That command will create `my-app` folder with 2 folders inside: 
+```
+my-app/
+ |- server
+ |- client
+```
+
+Folder `server` contains server application which is responsible for communication with external APIs and services.
```
Borales: @mateuszwozniak I believe it should be like "`server` folder ..."

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,74 @@
...
+npm start
+```
+and the client app in the second terminal:
+```bash
+cd my-app/client
 npm start
 ```
+then open http://localhost:3000 in your browser
+
+## Longer version
```
Borales: @mateuszwozniak `TL;DR` and `Longer version` - are not that different, can we "merge" them somehow?

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,74 @@
...
+cd my-app/client
+npm start # or yarn start
 ```
 
-or
+## Selecting the template for your project during generation
 
+Currently we provide just one template for the generator which delivers basic features of shop and blog (the template itself is called [shop-with-blog](https://github.com/deity-io/falcon/tree/master/examples/shop-with-blog).
+
+In the future we'll provide more templates that can be selected with `--example` option. Full command that uses `shop-with-blog`  in that case will be as following:
```
Borales: @mateuszwozniak should we mention `-h` option? it will show all available examples

packages/create-falcon-app/README.md
```diff
@@ -1,38 +1,74 @@
...
+npx create-falcon-app --example shop-with-blog my-app
 ```
 
-> Please note, some example templates use multi-application structure
-> (like `shop-with-blog` - it provides `client` and `server` sub-applications,
-> which have to be launched via separate terminal sessions). More specific
-> instructions will be provided at the end of the `create-falcon-app` call output.
+## Using your backend services
+
+By default `shop-with-blog` example uses demo Magento 2 and WordPress servers delivered by DEITY so you can check how DEITY Falcon works. If you cannot start the server part (from server folder) or you see any problems with server behavior please contact us via our [community slack channel](https://slack.deity.io).
```
Borales: @mateuszwozniak delivered -> provided? (as Deity does not deliver anything, but it does provide public instances)

packages/create-falcon-app/README.md
mateuszwozniak: Longer version is about everything below that header - explains how everything work etc. TL;DR is a few lines that make starting very easy if you don't want to dive into details.
So from my point of view those are 2 separate sections.

packages/create-falcon-app/README.md
mateuszwozniak: I don't think it matters in this case but I'll change it as you suggest.

PR #61, opened 2018-10-07
ci: install Yarn properly

DanielRuf:

Bug fix and docs update.

Yarn is installed with npm, which should not be the case.

Yarn is installed according to the official docs.

No.

**Other information:**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ node_js:
 cache: yarn
 
 before_install:
-  - npm i -g yarn
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.10.1
+  - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
   - yarn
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,7 +5,8 @@ node_js:
 cache: yarn
 
 before_install:
-  - npm i -g yarn
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.10.1
```
Borales: @DanielRuf I'm ok with your solution, but I don't think we need a specific version here. Maybe you have a specific concern about this?

DanielRuf: This is from https://yarnpkg.com/lang/en/docs/install-ci/

DanielRuf: Without the version flag it would install the latest release at this time.

Borales: @DanielRuf yes, good point. Thank you!

PR #77, opened 2018-10-11
Fix breadcrumbs

### LINKED ISSUE #62: Strange breadcrumb/link behaviour
**Describe the bug**
Some links are getting malformed under certain conditions which causes a blank page to show.

**To Reproduce**

1. On the sample shop, go to "Women" category
2. Click on "Tiffany Fitness Tee"
3. Click on "Tees" in breadcrumbs
4. Click on "Tiffany Fitness Tee"
5. Click on "Tops" in breadcrumbs (or any other breadcrumb link)

See how the URL is `http://localhost:3000/women/tops-women/women/tops-women/tees-women.html` instead of `http://localhost:3000/women/tops-women/tees-women.html`

**Expected behavior**
It should load the product page.

**Desktop (please complete the following information):**
 - OS: MacOS Sierra
 - Browser: Chrome
 - Version: 69

**Additional context**
~~Looks like the links are "appending" its href value to the current URL somehow, at first I though it was because the `to` prop in the links where missing the opening `/` but even after adding it the links were still pointing to the wrong URL even though it looks fine on the React Chrome tools.~~

After some digging the problem seems to be that links are using relative URLs instead of absolute.

mateuszwozniak:

Fixes #62 - urls in breadcrumbs are correctly rendered.

Urls in breadcrumbs were not prefixed with leading slash.

Leading slash has been added to urls of breadcrumbs items and last breadcrumb item is rendered as `Text` instead of `Link`.

No

### DIFF
--- a/packages/falcon-magento2-api/src/index.js
+++ b/packages/falcon-magento2-api/src/index.js
@@ -104,13 +104,13 @@ module.exports = class Magento2Api extends Magento2ApiBase {
   }
 
   /**
-   * Add suffix to path
+   * Add leading slash and suffix to path
    * @param {string} path - path to convert
    * @return {string} converted path
    * @todo get suffix from Magento2 config
    */
   convertPathToUrl(path) {
-    return `${path}.html`;
+    return path ? `/${path}.html` : path;
   }
 
   /**
--- a/packages/falcon-ecommerce-uikit/src/Breadcrumbs.tsx
+++ b/packages/falcon-ecommerce-uikit/src/Breadcrumbs.tsx
@@ -1,14 +1,18 @@
 import React from 'react';
 import { Link as RouterLink } from 'react-router-dom';
-import { Breadcrumbs as BreadcrumbsLayout, Breadcrumb, Link } from '@deity/falcon-ui';
+import { Breadcrumbs as BreadcrumbsLayout, Breadcrumb, Link, Text } from '@deity/falcon-ui';
 
 export const Breadcrumbs: React.SFC<{ breadcrumbs: any }> = ({ breadcrumbs }) => (
   <BreadcrumbsLayout>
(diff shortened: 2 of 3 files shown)


packages/falcon-magento2-api/src/index.js
```diff
@@ -104,13 +104,13 @@ module.exports = class Magento2Api extends Magento2ApiBase {
...
   /**
-   * Add suffix to path
+   * Add leading slash and suffix to path
    * @param {string} path - path to convert
    * @return {string} converted path
    * @todo get suffix from Magento2 config
    */
   convertPathToUrl(path) {
-    return `${path}.html`;
+    return path ? `/${path}.html` : path;
```
Borales: @sergy-gzh we need your help on this. Does Magento always use `.html` as URL suffix? is that an optional config?

Borales: @mateuszwozniak can we check something like `! path.startsWith('/')` too? just to avoid double-slash

mateuszwozniak: I was thinking about that, but as far as I know we always get paths without leading slash. Am I wrong?

sambolek: It's configurable via admin interface https://magento.stackexchange.com/questions/115194/magento-2-product-url-suffix

Borales: @sambolek thank you for a quick answer! 👍

mateuszwozniak: Thanks for the hint @sambolek. We're going to implement that in #79 and [falcon-magento2-module](https://github.com/deity-io/falcon-magento2-module/issues/5)

Borales: We decided to implement it properly once we'll be receiving this data from Magento endpoint, closing this conversation

PR #78, opened 2018-10-11
Google Analytics support

### LINKED ISSUE #75: Google Analytics
**What should be changed/improved?**
Introduce Google Analytics features

Borales:

Provides Google Analytics support

No

**Other information:**
Fixes #75


Borales:
@mateuszwozniak @tutok @tadeuszwojcik I'd appreciate hearing your initial thoughts on this changes before putting some (tricky) tests

Borales:
@tadeuszwojcik thank you for the constructive feedback 👍 
I've been using `react-ga` "knowledge" from a while ago, I didn't know they did so many improvements (like `testMode` and `debug` options) - changed that.

- it is still possible to use `react-ga` standalone, although our `@deity/falcon-analytics-plugin` provides a `withAnalytics` HOC that wraps `App` component and handles route change properly (and sends "pageview" data to Google Analytics); it also re-exports all available `react-ga` methods (just for the convenience)
- `react-with-analytics` removed as it uses "unsafe" react method, our package provides a proper one
- I added "Package naming conventions" section to the main README file
- 4.29kb with all dependencies, minified and gzipped

tadeuszwojcik:
Cool, thanks, that looks better. 
I'm still not sold on that plugin vs extension thing but that's for different conversation.

Also I'm currently not sure about the need to wrap `<App>` component in two HOCs -`withRouter(withAnalytics(App))`  I guess that pattern could get really hard to read & maintain when we'll potentially have more of those HOCs plugins, but maybe it's just me  not being a big fan of HOCs.

What if instad of `withAnalytics` we'd have `<TrackPageView/>` component that could simply be rendered inside the app.
It could be implemented in really similar way to `withAnalytics` HOC, something like:
```jsx
import React from 'react';
import ga from 'react-ga';
import { withRouter } from 'react-router-dom';

class TrackPageViewComponent extends React.PureComponent {
  componentDidMount() {
    const page = this.props.location.pathname;
    ga.pageview(page);
  }

  componentDidUpdate(prevProps) {
    const currentPage = this.props.location.pathname;
    const prevPage = prevProps.location.pathname;
    if (currentPage !== prevPage) {
      ga.pageview(currentPage);
    }
  }

  render() {
    return null;
  }
}

export const TrackPageView = withRouter(TrackPageViewComponent);
```
And then somewhere inside the `<App>` we could just  have `<TrackPageView/>` and that's it. 
Just react & it's lifecycle methods. This solution requires having `react-router-dom` as peer dependency inside the plugin, but anyway even currently it's sort of implicit dependency as well now as it won't work without react-router `withRouter` HOC.

BTW, I'm not sure from where `trackPage` comes from in your implementation, is it leftover after refactoring?.

Borales:
@tadeuszwojcik regarding 2 HOCs - I didn't want to put another dependency into this plugin's code (`react-router-dom`), but rather keeping it "clean" (trying to reduce the amount of "coupled packages"). 

It is possible to represent "TrackView" component as a router component, which should go before `<Switch>` block (so it could be rendered for every page), so in this case - we don't need `withRouter` method at all and having this - we don't need any other HOC:

```
const App = () => {
  ...
  <AppLayout>
    <Route path="/" component={TrackPage} />
    <Switch>
       ....
    </Switch>
  </AppLayout>
  ...
}
```

I'm aware of peer dependencies, but it would work without react-router (all it requires is `location.pathname` prop, that could be passed by anyone), so I'd insist on avoiding any react-router dependencies for this plugin (the more dependencies we add - the more problems we're gonna end up with). Maybe we could deliver both components and let the developer decide which one to use (some developers may find "decorator" pattern handier, so `withAnalytics` could help them with it).

`trackPage` comes from https://github.com/deity-io/falcon/blob/google-analytics/packages/falcon-analytics-plugin/src/index.ts#L8 (second arg), I didn't want to mix component with this the rendering logic (and plus it would ease the testing part).

tadeuszwojcik:
I generally agree about coupled packages but in our case we are literally adding package that glues together `react-ga` and `react-router-dom` to provide page tracking on route change. I don't see a reason why we should pretend that we don't. In fact maybe we should also have `react-ga` as peer dependency, as we are re-exporting it anyway. 

I don't find ` <Route path="/" component={TrackPage} />` to be most optimal API as I suppose that many people would wonder it it's supposed to track only `/` route.

All in all for me it still boils down to the case if it's even worth having that as separate package as it's literally less than 20 LOC that could also be placed simply in the shop & blog project.

Borales:
@tadeuszwojcik
> as I suppose that many people would wonder it it's supposed to track only `/` route.

we can say this about all other things too :) we **assume** that people know tools they work with, and hopefully, that won't be a case. But if that's the only concern - I wouldn't be worried too much over this.
 
`falcon-analytics-plugin` with "re-export" should provide us with an option to use it and its re-exported methods as a wrapper/proxy package of `react-ga`. Adding `react-ga` as a peer dep - we would force developers to add this package as their own dependency, so to enabled 1 feature - they would need to add 2 packages instead of 1? Makes no sense. 
The purpose of this package is to extend `react-ga` package. If they would need any of `react-ga` methods - they could have it by including this one package, if they don't need `withAnalytics` - they could easily **switch** to `react-ga` itself.

The reason why I don't add `withAnalytics` to examples is that I don't want to repeat myself and force other developers to do "copy-paste" from one project to another, or once React changes its API methods again - we would have to warn other developers to update their code from the example.
We shouldn't put too much code into our "boilerplate" example as it makes hardred to maintain the code in the future.

Borales:
I'm going to switch to `ganalytics` and move this package to `falcon-client`

Borales:
@tutok @mateuszwozniak  @tadeuszwojcik please do a code review

PR #83, opened 2018-10-18
improved error handling on invallid admin token

deity-bram:
throw a error when the admin token cant be found. probably the magento-api was not installed

### DIFF
--- a/packages/falcon-magento2-api/src/Magento2ApiBase.js
+++ b/packages/falcon-magento2-api/src/Magento2ApiBase.js
@@ -75,18 +75,20 @@ module.exports = class Magento2ApiBase extends ApiDataSource {
     Logger.info('Retrieving Magento token.');
 
     const response = await this.retrieveToken({ username: this.config.username, password: this.config.password });
-    
+
     // data available only if retrieveToken is not wrapped with cache.
     const tokenData = this.convertKeys(response.data || response);
     const { token, validTime } = tokenData;
-    if (token == undefined){
-        const noTokenError = new Error('Magento Admin token not found. did you install the falcon-magento2-module on magento?');
+    if (token === undefined) {
+      const noTokenError = new Error(
+        'Magento Admin token not found. Did you install the falcon-magento2-module on magento?'
+      );
 
-        noTokenError.statusCode = 501;
-        noTokenError.code = 'CUSTOMER_TOKEN_NOT_FOUND';
-        throw noTokenError;
-    } else{
-       Logger.info('Magento token found.');
+      noTokenError.statusCode = 501;
+      noTokenError.code = 'CUSTOMER_TOKEN_NOT_FOUND';
+      throw noTokenError;
+    } else {
+      Logger.info('Magento token found.');
     }
     this.token = token;
     this.tokenExpirationTime = null;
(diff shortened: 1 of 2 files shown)


packages/falcon-magento2-api/src/Magento2ApiBase.js
```diff
@@ -75,11 +75,19 @@ module.exports = class Magento2ApiBase extends ApiDataSource {
...
 
     const response = await this.retrieveToken({ username: this.config.username, password: this.config.password });
-
-    Logger.info('Magento token found.');
+    
     // data available only if retrieveToken is not wrapped with cache.
     const tokenData = this.convertKeys(response.data || response);
     const { token, validTime } = tokenData;
+    if (token == undefined){
+        const noTokenError = new Error('Magento Admin token not found. did you install the falcon-magento2-module on magento?');
```
Borales: @deity-bram as I remember - token comes from native Magento endpoint, not ours.

packages/falcon-magento2-api/src/Magento2ApiBase.js
```diff
@@ -75,11 +75,19 @@ module.exports = class Magento2ApiBase extends ApiDataSource {
...
+
+        noTokenError.statusCode = 501;
+        noTokenError.code = 'CUSTOMER_TOKEN_NOT_FOUND';
```
Borales: @deity-bram we have a dedicated `falcon-errors` meta-packages for such cases, we need to use it

deity-bram: Do you mean the package ' falcon-errors'? if so, I'll add my error and the one I copied from, because that one isn't in the package either. (did a grep to find the definition)

packages/falcon-magento2-api/src/Magento2ApiBase.js
deity-bram: apparently it doesn't. there is a different result if you enable/disable the magento module

packages/falcon-magento2-api/src/Magento2ApiBase.js
```diff
@@ -76,10 +76,20 @@ module.exports = class Magento2ApiBase extends ApiDataSource {
...
     // data available only if retrieveToken is not wrapped with cache.
     const tokenData = this.convertKeys(response.data || response);
     const { token, validTime } = tokenData;
+    if (token === undefined) {
+      const noTokenError = new Error(
+        'Magento Admin token not found. Did you install the falcon-magento2-module on magento?'
+      );
+
+      noTokenError.statusCode = 501;
+      noTokenError.code = 'CUSTOMER_TOKEN_NOT_FOUND';
```
Borales: @deity-bram you need to import `errors` pacakge and use one of its codes:

```
// on the top
import { codes } from '@deity/falcon-errors';

// on line 88
noTokenError.code = codes.CUSTOMER_TOKEN_NOT_FOUND
```

(

packages/falcon-magento2-api/src/Magento2ApiBase.js
```diff
@@ -367,7 +377,10 @@ module.exports = class Magento2ApiBase extends ApiDataSource {
    */
   createContextData(context) {
     if (!has(context, 'req.session')) {
-      throw new Error('No session in context passed to Magento2Api.createContextData()');
+      const noSessionError = new Error('No session in context passed to Magento2Api.createContextData()');
+      noSessionError.statusCode = 501;
+      noSessionError.code = 'SESSION_NOT_FOUND';
```
Borales: @deity-bram the same this with this error code

packages/falcon-magento2-api/src/Magento2ApiBase.js
```diff
@@ -11,6 +11,8 @@ const isEmpty = require('lodash/isEmpty');
 
 const DEFAULT_KEY = '*';
 
+import { codes } from '@deity/falcon-errors';
```
Borales: @deity-bram my mistake, I sorry, this package does not use es6, 

```suggestion
const { codes } = require('@deity/falcon-errors');
```

Borales: @deity-bram please accept this suggestion, otherwise this package will throw an error

PR #109, opened 2018-10-25
qucik fix for create-falcon-app

### LINKED ISSUE #105: React Context error
**Describe the bug**
Running an example project from `master` branch returns an error (`client`)

**To Reproduce**
1. Clone repo
2. Install dependencies for `example/shop-with-blog/client`
3. `yarn start`

**Screenshots**
image 1

tadeuszwojcik:
This is quick fix for preventing `npx create-falcon-app` crashing after react 16.6 release. 
I'll follow up with proper fix later on (so we are compatible with v16.6) or explanation why we can't be yet depending what I'll find out.

**Other information:**
Fixes #105


tadeuszwojcik:
@cane-lagerwaard `~` means only patch can be updated, if we'd go with <=16.6 we'd have to have 
something like `>=16.0 || <=16.6` as we don't support react 15 I think. From my perspective both version schemes (~ and ><) would quick fix that issue so `create-falcon-up` is usable again.

Borales:
@tadeuszwojcik please check Travis pipeline

mateuszwozniak:
Patch version is ok, we always want specified or newer patch, never previous one, so it's fine.

PR #114, opened 2018-10-29
Cart operations

### LINKED ISSUE #90: Cart operations
Implementation of cart operations:
- [ ] add to cart
- [ ] remove from cart 
- [ ] modify cart items 
- [ ] product options 
- [ ] error handling

mateuszwozniak:
This pull request introduces cart operations (fixes #90):
- adding products to cart (simple products, and products with configurable options)
- removing products from cart
- changing amount of items in the cart from minicart view

Folders and files structure follows our initial assumptions but I'm open for discussion about that.


packages/falcon-ecommerce-uikit/src/Header/Header.tsx
```diff
@@ -96,7 +96,11 @@ export const Searchbar = () => (
     </Link>
     <Icon gridArea={SearchBarArea.login} src="user" />
     <ToggleMiniCartMutation>
-      {toggle => <Icon gridArea={SearchBarArea.cart} src="cart" onClick={toggle as any} css={{ cursor: 'pointer' }} />}
+      {toggle => (
+        <MiniCartQuery>
+          {(data: any) => <MiniCartIcon onClick={toggle} gridArea={SearchBarArea.cart} itemsQty={data.cart.itemsQty} />}
```
Borales: @mateuszwozniak could we partially describe this data - `data.cart.itemsQty`? (since we expect at least this structure to be passed to `MiniCartIcon` component)

packages/falcon-ecommerce-uikit/src/Locale/LocaleContext.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+
+// todo: this is temporary, these values should be fetched from shop settings
+// todo: this probably should be exposed as LocaleProvider's prop so it can be passed from outside
+const GET_LOCALE_SETTINGS = gql`
+  query LocaleSettings {
```
Borales: @mateuszwozniak query names - upper case?

packages/falcon-ecommerce-uikit/src/MiniCart/MiniCartMutations.tsx
```diff
@@ -12,3 +12,51 @@ export class ToggleMiniCartMutation extends Mutation {
     mutation: TOGGLE_MINI_CART
   };
 }
+
+export const ADD_TO_CART_MUTATION = gql`
+  mutation addToCart($input: AddToCartInput!) {
```
Borales: @mateuszwozniak upper case

packages/falcon-ecommerce-uikit/src/Product/Product.tsx
```diff
@@ -0,0 +1,236 @@
+import React from 'react';
+import { Formik, Form, ErrorMessage } from 'formik';
+import { adopt } from 'react-adopt';
+import { themed, Box, Text, H1, NumberInput, Button, Icon, FlexLayout } from '@deity/falcon-ui';
+import { Breadcrumbs } from '../Breadcrumbs';
+// import { ProductMeta } from './ProductMeta';
```
Borales:
```suggestion
```

packages/falcon-ecommerce-uikit/src/MiniCart/MiniCartMutations.tsx
```diff
@@ -12,3 +12,51 @@ export class ToggleMiniCartMutation extends Mutation {
...
+
+export class RemoveCartItemMutation extends Mutation {
+  static defaultProps = {
+    mutation: REMOVE_CART_ITEM_MUTATION,
+    refetchQueries: ['miniCart', 'cart']
+  };
+}
+
+export const UPDATE_CART_ITEM = gql`
+  mutation updateCartItem($input: UpdateCartItemInput!) {
```
Borales:
```suggestion
  mutation UPDATE_CART_ITEM($input: UpdateCartItemInput!) {
```
?

examples/shop-with-blog/client/src/clientState.js
```diff
@@ -29,43 +29,46 @@ const sortOrders = [
...
-    src: 'https://picsum.photos/600/600?image=824',
-    name: 'Zoku Round Pop Moulds',
-    price: 199
-  },
-  {
-    src: 'https://picsum.photos/600/600?image=493',
-    name: 'Fusionbrands Poachpod Egg Poacher',
-    price: 299
+const MINI_CART_QUERY = gql`
+  query miniCart {
```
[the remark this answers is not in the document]
mateuszwozniak: No, that's not a correct approach imo. Uppercase names should be used for constants and query name should be either camel case (`miniCart`) or pascal case (`MiniCart`). In fact I think that pascal case would be the best option here. What's more, using constants would be a bit confusing for developers, as we're using also query names for refetching:
```
<Mutation mutation={ADD_TO_CART} refetchQueries={['miniCart']}>...</Mutation>
```

From my point of view using uppercase names would cause confusions here as there are 2 things we pass - const gql expression passed as `mutation` and query name passed as `refetchQueries`.

Also, official GraphQL resources use pascal case (https://graphql.org/learn/queries/).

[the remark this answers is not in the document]
Borales: @mateuszwozniak what I meant is that we should stick to some sort of convention and follow it across our packages. I am ok with either way, we just need to adopt one of them (avoid any confusion for the rest of community devs)

packages/falcon-ecommerce-uikit/src/Header/Header.tsx
Borales: @mateuszwozniak we can skip this comment for now

examples/shop-with-blog/client/src/clientState.js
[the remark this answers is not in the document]
mateuszwozniak: Yep, I know :) I'll adjust other queries to pascal case so everything's consistent.

packages/falcon-ecommerce-uikit/src/Product/Product.tsx
```diff
@@ -0,0 +1,234 @@
...
+            <ProductGallery items={product.gallery} />
+          </Box>
+          <Text fontSize="sm" gridArea={Area.sku}>
+            {`${translations.sku}: ${product.sku}`}
+          </Text>
+          <H1 gridArea={Area.title}>{product.name}</H1>
+          <ProductForm
+            sku={product.sku}
+            validate={this.createValidator(product)}
+            render={this.renderProductFormContent}
```
tutok: can we render product form content as children for `ProductForm`? Then we could have something like this:

```
 <ProductForm  sku={product.sku} validate={this.createValidator(product)}>
   {({ addToCartMutation, formik: { values, isSubmitting, errors, setFieldValue, submitCount }, productConfigurator }) => (
     // here content of  `this.renderProductFormContent` passed inline
   )}
</ProductForm>
```

then our code wll be more `react`ish. 

Second thing is thatcontent of `this.renderProductFormContent` could be extracted as separate component.

packages/falcon-ecommerce-uikit/src/Product/Product.tsx
```diff
@@ -0,0 +1,234 @@
...
+    addToCartMutation,
+    formik: { values, isSubmitting, errors, setFieldValue, submitCount },
+    productConfigurator
+  }: any) => {
+    const { product, translations } = this.props;
+
+    return (
+      <React.Fragment>
+        <Price fontSize="xxl" gridArea={Area.price} value={product.price} />
+        <Form>
```
tutok: Can `Form` component be a part of  `<ProductForm>` component? 
lets say [here](https://github.com/deity-io/falcon/pull/114/files?utf8=%E2%9C%93&diff=split#diff-34a94b1e75480691ea522c475dc54ac7R134) we could have

```
 <ProductConfigurator {...someProps}  >
    <Form>
      {render}
    <Form>
 </ProductConfigurator>
```

Right now we have  `<ProductForm>` where its name sugest that html form is rendered inside while it is not.

mateuszwozniak: Good point. I can move `<Form>` to formik component and that way it will not be required here. Will change that.

packages/falcon-ecommerce-uikit/src/Product/Product.tsx
mateuszwozniak: To be honest I wanted to make the code as readable as possible here. Content of the form is quite complex and passing `this.renderProductFormContent` as `render` prop seemed most readable for me (and that's another way of using render props).

When it comes to creating a separate component - I prefer to leave it as it is now, as it's easier to read - you have everything on one screen size even if that is extracted. Also I'm not sure if moving it outside as e.g. `<ProductFormContent/>` would give us anything?

packages/falcon-ecommerce-uikit/src/MiniCart/MiniCart.tsx
```diff
@@ -34,57 +39,99 @@ const miniCartProductTheme: DefaultThemeProps = {
...
+            src="remove"
+            mr="sm"
+            css={({ theme }) => ({
+              ':hover': { stroke: theme.colors.secondary }
+            })}
+          />
+        </Link>
+      )}
+    </RemoveCartItemMutation>
+    <UpdateCartItemMutation>
```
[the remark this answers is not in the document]
mateuszwozniak: Yes, we don't have base mutation. I'm still thinking about that, but I'd rather wait for changes from Apollo to their libraries as I'm not sure if we won't have to re-work our Query after their updates :(

packages/falcon-ecommerce-uikit/src/Cart/CartQuery.tsx
```diff
@@ -0,0 +1,36 @@
...
+    open: boolean;
+  };
+  cart: {
+    itemsQty: number;
+    quoteCurrency: string;
+    items: any[];
+  };
+};
+
+export class CartQuery extends Query<CartData> {
```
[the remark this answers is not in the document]
mateuszwozniak: Nope, because we can get `CartData` not only from the query, but also from mutation (or even create it by ourselves).

packages/falcon-ecommerce-uikit/src/Product/Product.tsx
mateuszwozniak: Done :)

packages/falcon-ecommerce-uikit/src/Cart/CartQuery.tsx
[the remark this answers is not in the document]
tutok: But will we always also fetch minicart?
```
  miniCart: {
    open: boolean;
  };
```

I think we could have many cart queries which return different fields, as it depends on context

packages/falcon-ecommerce-uikit/src/MiniCart/MiniCart.tsx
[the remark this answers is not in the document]
tutok: sure, I want to just be sure that we are on the same page.

packages/falcon-ecommerce-uikit/src/Product/Product.tsx
tutok: What I want to say is I am not sure if the component injection is good (consistent with other codebase) here, simply I would prefer to render entire form inline.

We were agreed that we split our components into two groups: 
1. Query/Mutation components - which are responsible for data reading and manipulation (kind of API for second group)
1. and all Visual (the rest) components which used data provided by the Query/Mutation components.
 
And having in mind that `ProductForm` actually wraps `ToggleMiniCartMutation` and `AddToCartMutation` mutations (and ads `form` specifics) then for me it becomes a member of 
Query/Mutation components. Then according to our rule that we do not inject Visual components into Query/Mutation components I would simply render `<ProductFormContent/>` as a children of `<ProuctForm>`. As we should have the same rules everywhere to be consistent.

mateuszwozniak: It's indeed component injection but it's also a render prop approach :) So the only acceptable change for me is placing this component inline as you mentioned in the first comment, as I don't want to extract that to a separate component just for the sake of having it as a "Visual" component. 
I'll do the change in a few minutes.

packages/falcon-ecommerce-uikit/src/Cart/CartQuery.tsx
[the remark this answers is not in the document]
mateuszwozniak: Actually, this one is a very nice catch. We won't have miniCart in this query - that's my copy/paste result. Here, we'll have only cart data required for displaying full cart view - Mini Cart is in `MiniCartQuery.tsx`

packages/falcon-ecommerce-uikit/src/Product/Product.tsx
mateuszwozniak: Done

examples/shop-with-blog/client/src/App.js
```diff
@@ -43,33 +44,35 @@ const Category = AsyncComponent(() => import(/* webpackChunkName: "shop/category
...
-          </AppLayout>
-        </ThemeProvider>
-        <ThemeEditor {...props} />
-      </React.Fragment>
-    )}
-  </ThemeState>
+  <LocaleProvider>
+    <ThemeState initial={deityGreenTheme}>
+      {props => (
+        <React.Fragment>
```
cane-lagerwaard: why do we use `<React.Fragment>` and not `<Fragment>`? Simply use `import { Fragment }  from react;`

mateuszwozniak: No particular reason for that. It's probably just my preference as when I see `React.Fragment` I immediately know what's that. We could also use `<>` and `</>`. Do you want me to change it?

cane-lagerwaard: `<>` looks clean, but I'm not sure about support for this: https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html#support-for-fragment-syntax if not supported, just leave it as is right now as I agree with you that it is immediately clear

mateuszwozniak: Ok, I'll take a look at support in babel maybe next week, as I'm going to update various libraries that canno be done automatically by renovate.

PR #134, opened 2018-11-04
Codecov

Borales:
Providing Codecov supprt (https://codecov.io/gh/deity-io/falcon)


tadeuszwojcik:
Bit off topic, but is there a particular reason why we are using feature tag along with `feat` messages for build, tests and docs related stuff?

Issue #136, opened 2018-11-05
shopping cart UI improvements


tadeuszwojcik:
- [x] add empty state (Shopping basket is empty message or similar)
- [x] improve layout when basket items overflow viewport height
- [x] optimistically update UI when deleting items from the cart (currently it's waiting server response)?
- [x] disable "Add to cart" button when adding to cart is in progress
- [x] after "Add to cart" click show mini-cart when is already loaded with content
- [x] close button position tweaks so it aligns better with the design
- [x] localize messages

PR #143, opened 2018-11-08
Shopping cart ui improvements

tadeuszwojcik:

**Other information:**
closes https://github.com/deity-io/falcon/issues/136


packages/falcon-ecommerce-uikit/src/Cart/CartMutation.tsx
```diff
@@ -17,20 +18,43 @@ export const ADD_TO_CART = gql`
...
+        data: {
+          removeCartItem: { itemId }
+        }
+      }: any
+    ) => {
+      const data = store.readQuery({
+        query: GET_CART
+      });
+
+      data.cart.items = data.cart.items.filter((item: any) => item.itemId !== itemId);
```
Borales: @tadeuszwojcik `refetchQueries: ['Cart']` - doesn't it fetch the cart data from the remote backend already? or is it done for "optimistic" response?

tadeuszwojcik: There are two queries, `Cart` and `Minicart`. It used to be that delete wasn't `optimistic` so user had to wait until server responds with refetch for `Minicart` now it's handled optimistically and deleting from mini cart is instant for end user...but we still need to refetch `Cart` query (for full Cart page) as it has things like total price, discounts etc which are calculated on the backend.

packages/falcon-ecommerce-uikit/src/MiniCart/MiniCart.tsx
```diff
@@ -47,6 +47,31 @@ const miniCartProductTheme: DefaultThemeProps = {
...
+const miniCartLayout: DefaultThemeProps = {
+  miniCartLayout: {
+    display: 'grid',
+    gridRowGap: 'md',
+
+    // prettier-ignore
+    gridTemplate: toGridTemplate([
+      ['1fr',                    '30px'                         ],
+      [MiniCartLayoutArea.title, MiniCartLayoutArea.close,      ],
+      [MiniCartLayoutArea.items, MiniCartLayoutArea.items, '1fr'],
```
mateuszwozniak: What does that additional `'1fr'` at the end of this row?

tadeuszwojcik: It's row height, by default each row height is set to 'auto' implicitly, but here we want items row to be as high as possible (so checkout button is always at the bottom and title at the top).
There's also additional code that makes items row scrollable so even if there are many items in minicart checkout and title are always visible and only items area scrolls, does it make sense?

mateuszwozniak: Yes, clear. Thanks for the explanation :)

PR #164, opened 2018-11-16
Fix ssr

tutok:
Because of `react-apollo` in version `2.3.1` introduces changes in tree walking, and our other dependencies fail. We need to specify the fixed version of it into `2.2.4`


.gitignore
```diff
@@ -3,10 +3,8 @@ node_modules/
 coverage/
 .DS_Store
 .idea/
-yarn.lock
 yarn-error.log
 package-lock.json
-yarn-error.log
```
Borales: @tutok please revert this

tutok: I removed duplicated `yarn-error.log` record, nothing to rever

Issue #157, opened 2018-11-13
Sidebars handling improvements


tadeuszwojcik:
- [x] use single sidebar 'portal' instance for all the sidebars
- [x] generalize mutations & queries for various sidebars to avoid unnecessary repetition
- [x] close button

- [x] lazy load components that are rendered inside sidebars? (that would improve time to interactive hopefully, needs bit more investigation)
- [x] (how to) specify sidebar side given we'll have single sidebar instance (needs bit more investigation/research)

PR #172, opened 2018-11-20
Sidebars improvements

tadeuszwojcik:

**Other information:**
Closes https://github.com/deity-io/falcon/issues/157


packages/falcon-ecommerce-uikit/src/Sidebar/SidebarMutations.tsx
```diff
@@ -0,0 +1,31 @@
...
+`;
+
+export class CloseSidebarMutation extends Mutation {
+  static defaultProps = {
+    mutation: CLOSE_SIDEBAR_MUTATION
+  };
+}
+
+type OpenSidebarVariables = {
+  contentType: 'cart' | 'signin';
```
tutok: should not be a `string`? What if I would like to add to my shop mini dashboard sidebar, then I need to modify ecommerce-uikit?

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { Loader, CloseSidebarMutation } from '@deity/falcon-ecommerce-uikit';
+import { Sidebar as SidebarLayout, Portal, Backdrop, Icon, Box } from '@deity/falcon-ui';
+import AsyncComponent from 'src/components/Async';
+import { SidebarQuery } from './index';
+
+const SidebarContents = AsyncComponent(() =>
+  import(/* webpackChunkName: "shop/SidebarContents" */ './SidebarContents')
```
tutok: maybe each 'sidebar content' can be async and individually loaded, instead of fetching all possible sidebar contents at one? Will it have huge performance drawback?

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,64 @@
...
+);
+
+export class Sidebar extends React.Component {
```
tutok: I am thinking if it will be possible to extract `SidebarLogic` (to be consistent with names, but I think SidebarHost could be better) component  which will cover all internals like `SidebarQuery`, `CloseSidebarMutation` and `componentDidMount()`

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,64 @@
...
+  render() {
+    return (
+      <SidebarQuery>
+        {({ sidebar }) => (
+          <CloseSidebarMutation>
+            {closeSidebar => (
+              <React.Fragment>
+                <SidebarLayout as={Portal} visible={sidebar.open} side={sidebar.side || undefined}>
+                  <Box position="relative" flex={1}>
+                    <Icon
```
tutok: what do you think about moving X icon into SidebarLayout? on the end, we could do:

```
<SidebarLogic>
{(props) => 
    <SidebarLayout {...props}>
        <SidebarContent .... />
    </SidebarLayout>
}
</SidebarLogic>
```

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,64 @@
...
+                      src="close"
+                      stroke="black"
+                      position="absolute"
+                      right={0}
+                      top={0}
+                      onClick={() => closeSidebar()}
+                    />
+                    {this.state.ready ? <SidebarContents contentType={sidebar.contentType} /> : <Loader />}
```
tutok: will it be possible to render `SidebarContent`s individually? 

```
<SidebarLayout {...props}>
        <SidebarContent1 type="sidebarContent-1" .... />
        <SidebarContent2 type="sidebarContent-2" .... />
        <SidebarContent2 type="sidebarContent-3" .... />
    </SidebarLayout>
```
and `Loader` component could be rendered via `SidebarLayout`?

examples/shop-with-blog/client/src/clientState.js
```diff
@@ -29,20 +29,21 @@ const sortOrders = [
   }
 ];
 
-const GET_MINI_CART_STATE = gql`
-  query MiniCartState {
-    miniCart @client {
+const GET_SIDEBAR_STATE = gql`
+  query SIDEBAR_STATE {
```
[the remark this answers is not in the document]
tadeuszwojcik: Good point, I've tried that before, somehow didn't work, but tried now and it's fine.

packages/falcon-ecommerce-uikit/src/Sidebar/SidebarMutations.tsx
tadeuszwojcik: Those mutations are only temporarily in ecommerce-ukit, it's unfortunate, but we'll move them (at least local state ones) to shop-and-blog project when moving rest of the ecommerce-uikit ui related  stuff.

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
tadeuszwojcik: What I think we could have is (it's a full version, we could spread props as well, but wanted to show explicitly what we need to pass around):
```jsx
     <SidebarContainer>
        {props => (
          <Sidebar
            ready={props.ready}
            onClose={props.closeSidebar}
            visible={props.visible}
            side={props.side}
            contentType={sidebar.contentType}
            content={SidebarContents}
          />
        )}
      </SidebarContainer>
```
I'm open to reworking it in this way so we have `SidebarContainer` (IMHO better than host or logic, but open to change that).

Instead of `SidebarLayout` I'd make it simply `Sidebar` (so it has layout close icon and loader inside).

As you can see I've passed `SidebarContents` as content prop, as if we'd render it right away as `<SidebarContent/>` it would make async component loading trigger immediately and I wanted to delay that. We could also use render props:
```jsx
<SidebarContainer>
        {props => (
          <Sidebar
            ready={props.ready}
            onClose={props.closeSidebar}
            visible={props.visible}
            side={props.side}
          >
            {() => <SidebarContents contentType={contentType} />}
          </Sidebar>
        )}
</SidebarContainer>
```
The downside is that we'll have little bit of nesting, but it doesn't look that bad, so perhaps that is the better way.

@tutok , @mateuszwozniak  please let me know what do you think.

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
tadeuszwojcik: (resolving as covered in comments already)

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
tadeuszwojcik: `SidebarContents` is a simple convenience wrapper around other pretty high level components so if someone would like to render let's say mini cart it would be just ` <MiniCartQuery>{data => <MiniCart {...data} />}</MiniCartQuery>` . In this wrapper I render all the contents but with `hidden` attribute where appropriate to hint react that it's low priority rendering, but it still renders so there is pretty good change that when user wants to see it it's already there (no animation junk etc).
I could split those to look like:
```jsx
    <SidebarContainer>
        {props => (
          <Sidebar ready={props.ready} onClose={props.closeSidebar} visible={props.visible} side={props.side}>
            {contentType => (
              <>
                <SidebarContent active={contentType === 'cart'}>
                  <MiniCartQuery>{data => <MiniCart {...data} />}</MiniCartQuery>
                </SidebarContent>
                <SidebarContent active={contentType === 'signin'}>
                  <MiniAccountQuery>{data => <MiniAccount {...data} />}</MiniAccountQuery>
                </SidebarContent>
              </>
            )}
          </Sidebar>
        )}
      </SidebarContainer>
```

but then how I could load whole sidebar render prop contenents(one with many `SidebarContent `) using async component? It needs to be provided with single component hence wrapper `SidebarContents`. Does it make sense?

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
tutok: I see, you are right. What in my opinion looks weird is that we need to have `SidebarContents` component which explicitly needs to know what kind of components can render. What I try to suggest is to do it somehow configurable. But then we loose feature async feature...

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
tutok: First of all, please forgive me that I spread my thoughts (after initial review), between many comments on different files. There was the single idea behind it, but I believe you realize it, Moreover, I clearly understand where this idea fails :P. 

I think the above example is a really nice compromise.

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
tutok: you are right loading asynchronously each possible sidebar content does not make snese...

packages/falcon-ecommerce-uikit/src/Sidebar/SidebarMutations.tsx
tutok: I am confused, `'cart' | 'signin'` is not equal to `string`

tadeuszwojcik: OK, I've changed it to string. That union type was only as a convenience to not make mistakes writing non existing content type  as long as it's in uikit package.

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/Sidebar.js
tadeuszwojcik: Done.

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/SidebarContainer.js
```diff
@@ -0,0 +1,54 @@
...
+  render() {
+    return (
+      <SidebarQuery>
+        {({ sidebar }) => (
+          <CloseSidebarMutation>
+            {closeSidebar =>
+              this.props.children({
+                ready: this.state.ready,
+                close: closeSidebar,
+                open: sidebar.open,
```
tutok: for me it is quite hard to realize at first glance if `close` / `open` is `func` / `bool`, I would change `open` to `isOpen` (and we could be consistent `sidebar.open` -> `sidebar.isOpen`)

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/SidebarContents.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Box } from '@deity/falcon-ui';
+import { MiniCartQuery, MiniAccountQuery, MiniCart, MiniAccount } from '@deity/falcon-ecommerce-uikit';
+
+export default ({ contentType }) => {
+  if (!contentType) return null;
```
tutok: should not we return `null` also in the case when `contentType` is not recognized, let say `test` in this case?

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/SidebarContainer.js
tadeuszwojcik: Done.

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/SidebarContents.js
tadeuszwojcik: I've added proper error/warning for case when unrecognized content type is provided.

PR #173, opened 2018-11-20
Apollo config

### LINKED ISSUE #166: Pass apollo client configuration into ApolloClient explicitly
`falcon-client/src/service/ApolloClient` extract apollo client configration from `clientState` or `initial-state` which is taken from `window.__APOLLO_STATE__`.
This configuration setting should be passed explicitly, ApolloClient function should not be aware of initialState which can be varying.

Moreover, it is not possible to create an instance of ApolloClient without configuration, so we should not even check if it's provided, it needs to be. Then we can use the pattern of `bootstrap.js` which provide all defaults for `falcon-client` and put there all necessary ApolloClient configuration.

method `expandValue` located in ApolloClient.js should not be there and name should be also changed as it does not describe what it is doing.

tutok:
Passing Apollo Client configuration explicite. closes #166


packages/falcon-client/src/service/ApolloClient.js
```diff
@@ -1,85 +1,61 @@
...
+    headers,
+    apolloClientConfig
+  } = config;
+  const { httpLink, connectToDevTools, ...restApolloClientConfig } = apolloClientConfig;
+  const addTypename = false; // disabling 'addTypename' option to avoid manual setting "__typename" field
 
-  // disabling 'addTypename' option to avoid manual setting "__typename" field
-  const addTypename = false;
   const cache = new InMemoryCache({ addTypename }).restore(initialState);
-  const linkState = withClientState({
```
Borales: @tutok could you please revert changes for these `linkState` and `httpLink` consts? I defined them separately just to avoid deep nesting calls

tutok: I can not rever const `linkState` or `httpLink` as these names are used by config object. When I remove them, then I can also remove all that wierd and overused there syntax: `const { linkState = {}`

tutok: I can not rever `linkState` or `httpLink` consts as these names are already used. After removing them I can simplify overused syntax: `const { linkState = {}: linkStateConfig } = apolloClientConfig;` which is harder to understad IMHO.

Borales: @tutok ok, not literally, you could use different names (like `linkStateInstance`, `httpLinkInstance` etc)

tutok: sure, I knew, done. But personally still does not see any reason to do that.

Issue #190, opened 2018-11-27
Account recovery flow


tadeuszwojcik:
- [ ] forgot password form linked from sign in
- [ ] password reset form (opened via email link)

PR #195, opened 2018-11-28
Account recovery

tadeuszwojcik:

**Other information:**
Closes https://github.com/deity-io/falcon/issues/190


packages/falcon-ecommerce-uikit/src/AccountRecovery/ResetPassword.tsx
```diff
@@ -0,0 +1,132 @@
...
+        </GridLayout>
+      );
+    }}
+  </ValidatePasswordTokenQuery>
+);
+
+export const InvalidToken: React.SFC = () => (
+  <OpenSidebarMutation>
+    {openSidebar => (
+      <>
```
mateuszwozniak: Please use `<React.Fragment>` as we do in other places to keep things consistent.

packages/falcon-ecommerce-uikit/src/AccountRecovery/ResetPassword.tsx
```diff
@@ -0,0 +1,132 @@
...
+                  label="New Password"
+                  name="password"
+                  required
+                  type="password"
+                  autoComplete="new-password"
+                >
+                  {inputProps => <PasswordRevealInput {...inputProps} />}
+                </FormField>
+
+                <FormSubmit submitting={loading} value="Reset my password" />
```
mateuszwozniak: Can we center that button? It looks a bit weird when right-aligned, especially when "loading" animation starts.

packages/falcon-ecommerce-uikit/src/Forms/FormField.tsx
```diff
@@ -97,20 +98,24 @@ export const FormField: React.SFC<FormFieldProps> = ({
         const error = getIn(form.errors, name);
         const invalid = !!touch && !!error;
 
+        // TODO: is there a better way of handling input ids (more automated)?
+        // is fallback to name correct?
```
mateuszwozniak: It's correct as long as we have 1 field with a particular name on the whole page, which is not true e.g. for the checkout form, where we have 2 address forms rendered. We could probably use some hack at `Form` level, e.g. force users to pass id to the form and generate id for the fields from form's id and field's name to avoid forcing ids on all the fields.

packages/falcon-ecommerce-uikit/src/AccountRecovery/ResetPassword.tsx
tadeuszwojcik: Done.

packages/falcon-ecommerce-uikit/src/Forms/FormField.tsx
tadeuszwojcik: I was thinking about it, but we'd have to introduce another context for `Form` component in addition to formik's one and use it in `FormField`. I wasn't convinced about that, hence currently one can specify `id` explicitly on `FormField`.

PR #212, opened 2018-12-04
Ejecting ecommerce-uikit package

tadeuszwojcik:

**Other information:**

It can be tested by running `yarn eject` command inside shop-with-blog client dir.
More description about this approach will be delivered as separated docs/RFC later on, but even before that it would be great to have it available in nightly builds to test it out how it works outside monorepo.


CHANGELOG.md
```diff
@@ -27,9 +27,14 @@ Versions marked with a number and date (e.g. Falcon Client v0.1.0 (2018-10-05))
 
 ### Create Falcon App v1.0.7 (2018-10-25)
 
+- feat: allow ejecting `falcon-ecommerce-uikit` package via `eject` command (https://github.com/deity-io/falcon/pull/212))
```
Borales: @tadeuszwojcik does this relate to CFA?

packages/falcon-dev-tools/eject/README.md
```diff
@@ -0,0 +1,26 @@
...
+or with yarn:
+
+```bash
+yarn add --dev @deity/falcon-eject
+```
+
+## Usage
+
+```bash
+npx eject @deity/falcon-ecommerce-uikit src/uikit
```
Borales:
```suggestion
npx @deity/falcon-eject @deity/falcon-ecommerce-uikit src/uikit
```
?

packages/falcon-dev-tools/eject/README.md
```diff
@@ -0,0 +1,26 @@
+# Falcon Eject
+
+`@deity/falcon-eject` package is a CLI tool exposing single `eject` command that can eject package written in TypeScript,
+by converting it into JavaScript while preserving it's original formatting. It also replaces imports with ejected folder path and installs ejected package dependencies into the project where it's ejecting.
```
Borales:
```suggestion
by converting it into JavaScript while preserving its original formatting. It also replaces imports with the ejected folder path and installs ejected package dependencies into the project where it's ejecting.
```

packages/falcon-dev-tools/eject/eject.js
```diff
@@ -0,0 +1,179 @@
...
+      )
+    );
+  }
+
+  console.log(chalk.green(`${bold(packageToEject)} package ejected!`));
+}
+
+run();
+
+function convertToJs(ejectedFilesMeta) {
```
Borales: @tadeuszwojcik wouldn't it be more convenient to have a "dist-js" (or similar name) folder with JS files with TypeScript stripped down (like we discussed previously), so any developer could take a peek at any component he needs and whenever he wants?

- we could simply copy-paste those files without running the whole TS compilation
- a developer doesn't need to install TypeScript and a bunch of Babel plugins
- even if a developer already ejected uikit, he still may want to update his component from such `dist-js` folder

packages/falcon-dev-tools/eject/README.md
tadeuszwojcik: Sample usage section is based on the assumption that someone has already installed it, that's why npx eject

CHANGELOG.md
tadeuszwojcik: where to put it in that case? there is already 'fixed problem with React 16.6.0' that also affects example more, but we don't have example/shop-with-blog section

packages/falcon-dev-tools/eject/README.md
Borales: @tadeuszwojcik my bad, I didn't know it checks `node_modules/.bin` folder first

packages/falcon-dev-tools/eject/eject.js
tadeuszwojcik: And even if we won't decide to  go with ejecting packages workflow, that piece of code that transforms TS to JS will be useful anyway.

Borales: @tadeuszwojcik 
> I don't like workflow when one needs to looks into node_modules/dist/dist-js folder and look around what to copy&paste, but perhaps it's only me, maybe it won't be an issue for others.

My comment was referencing the implementation part only (in the context of Falcon packages), I like the idea itself. Once somebody ejects uikit - they are stuck with that state of the package. With every new version of uikit - these developer won't get the updates we provide them with. Having "dist-js" folder - yes, developer would need to make some efforts and find the required file, but still give him such ability would be nice. And we still can add required depedencies to the target project with "eject". But we can figure something out with the next version.

But speaking more "globally" (taking a step back from "falcon" context) - would it be better to name this package without "falcon" part? The reason why I assumed this will work with our packages only is the name of this package, probably it will make the same effect to other developers too (assuming this is a package that is being used as an internal dev tool for Falcon). What do you think? @mateuszwozniak @tutok

packages/falcon-dev-tools/eject/eject.js
```diff
@@ -0,0 +1,179 @@
...
+  const resolvedPackagePackageJsonPath = path.join(resolvedPackageDir, '../package.json');
+  const resolvedPackagePackageJsonDependencies = require(resolvedPackagePackageJsonPath).dependencies;
+  const packageSrcFiles = path.join(resolvedPackageDir, '../src');
+
+  console.log(chalk.green(`Ejecting ${bold(packageToEject)} package into ${bold(targetDirFullPath)} ...`));
+
+  // don't run any further if targetDir already exists, we don't want to overwrite anything
+  if (fs.pathExistsSync(targetDirFullPath)) {
+    console.log(
+      chalk.red(`Looks like ${bold(packageToEject)} has been already ejected ${bold(targetDir)} dir already exists.`)
```
mateuszwozniak:
```suggestion
      chalk.red(`Looks like ${bold(packageToEject)} has been already ejected because ${bold(targetDir)} dir already exists.`)
```

packages/falcon-dev-tools/eject/eject.js
```diff
@@ -0,0 +1,179 @@
...
+  // and then replace imports inside them
+  replaceImports(appFilesMeta, packageToEject, targetDir);
+
+  // as a final step install dependencies of package being ejected to project where it's ejected
+  // this logic fails when using yarn workspaces - https://github.com/yarnpkg/yarn/issues/4812
+  const dependenciesToInstall = Object.keys(resolvedPackagePackageJsonDependencies).map(
+    packageName => `${packageName}@${resolvedPackagePackageJsonDependencies[packageName]}`
+  );
+
+  console.log(chalk.green(`Installing ${bold(packageToEject)} package dependencies in the project...`));
```
mateuszwozniak:
```suggestion
  console.log(chalk.green(`Installing dependencies of ${bold(packageToEject)} package in the project...`));
```

packages/falcon-dev-tools/eject/eject.js
Borales: @mateuszwozniak `eject-pkg` is free, should be good I guess.

tadeuszwojcik: All right, so I guess please just let me know final name of the package, and that would be all for now.
Actually Alex I think (it's something we've partially discussed already if I recall correctly) it would be useful to have online docs for ecommerce-uikit similar to falcon-ui, where devs could see how component looks/behaves, props that accepts but also it's JS source code (ideally formatted via the same logic as eject package has), that would be the best way to surface the issue you've mentioned IMHO.

Borales: @tadeuszwojcik oh right, we have, that would be a perfect solution and a great place to refer to for other developers, I agree.

Package name - maybe `eject-ts`? @mateuszwozniak @tadeuszwojcik 
Just to be sure package name describes its purpose.

tadeuszwojcik: I like `eject-ts` better, but that would be the top level name, that's the assumption? not inside `@deity/` scope ?

Borales: @tadeuszwojcik right, unscoped

packages/falcon-dev-tools/eject/eject.js
```diff
@@ -0,0 +1,179 @@
...
+const fs = require('fs-extra');
+const klawSync = require('klaw-sync');
+const chalk = require('chalk');
+const babel = require('@babel/core');
+const recast = require('recast');
+const prettier = require('prettier');
+const install = require('install-packages');
+
+// as an improvement those options could be provided via command line perhaps
+const prettierOptions = {
```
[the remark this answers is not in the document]
tadeuszwojcik: I could do that for part of the config (still need to add parser option at least and line endings perhaps) , but I'm not sure about adding `@deity/eslint-config-falcon` as a dependency...

packages/falcon-dev-tools/eject/README.md
tadeuszwojcik: `npx` allows running CLI scripts straight from `node_modules/bin` , you could do something like that via npm, but it would need to define something in package.json `scripts` first (actually it's what I did for our example)

packages/falcon-dev-tools/eject/eject.js
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env node
```
Borales: @tadeuszwojcik could we move this file to `/eject/bin/eject.js`? (just to follow common practice for NPM bin files)

packages/falcon-dev-tools/eject/eject.js
```diff
@@ -0,0 +1,179 @@
...
+      plugins: [
+        '@babel/plugin-syntax-jsx',
+        '@babel/plugin-syntax-dynamic-import',
+        '@babel/plugin-syntax-class-properties',
+        ['babel-plugin-transform-rename-import', { original: originalImport, replacement: replacementImport }],
+        pluginRecast
+      ]
+    });
+
+    const transformedContents = prettier.format(result.code, prettierOptions);
```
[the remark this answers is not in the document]
tadeuszwojcik: Not sure if I understand you correctly, right now we run both prettier and line endings normalization  two times, once for transformed TS files from ejected package (after those are transformed) and second time for `app` files where we've updated imports.

And you'd like to do this 'prettier' step once after all transformations for all the files? 
I didn't want to do that this way I guess because `convertToJs` function could be reused just for converting TS->JS, but can change that if you want, doesn't change much in practice.

packages/falcon-dev-tools/eject/eject.js
tadeuszwojcik: My understanding was that bin dir is common practice when package has more files or also exports some code not only CLI, but sure, if you want I can change that and put it in `bin`

tadeuszwojcik: Done. good refactoring, will also make potential programmatic usage in the future  easier.

packages/falcon-dev-tools/eject/eject.js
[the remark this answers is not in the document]
tutok: thanks, I see, sounds reasonable.

packages/falcon-dev-tools/eject/eject.js
cane-lagerwaard: @tadeuszwojcik @Borales `eject-ts` seems fine, however I would still scope it - to notify users it's an official package from @deity

cane-lagerwaard: @tadeuszwojcik @Borales on second thought, it makes sense to have it unscoped - considering it's universal

tadeuszwojcik: Done.

Issue #112, opened 2018-10-25
React 16.6 compatibility


tadeuszwojcik:
Latest react update changed couple [internal things](https://github.com/facebook/react/pull/13829) related to Context handling and that currently breaks :
- react-apollo - https://github.com/apollographql/react-apollo/issues/2530
- [react-async-bootstrapper]( https://github.com/ctrlplusb/react-async-bootstrapper)

Both of those libs use `tree-walking` technique to "render react outside react" (to gather info about queries to run etc) that is very fragile and breaks almost every time react gets updated.

I can most likely fix `react-async-bootstrapper` as forked it already and did few fixes before, but not sure react-apollo.

tadeuszwojcik:
I've looked into this issue a little bit more.
Fixing react-apollo and async-bootsrapper will help with v16.6, but will break again with v16.7 and react hooks  (already confirmed https://github.com/apollographql/react-apollo/issues/2541). 

Fundamental approach taken by those libs ("react tree walking") has many flaws and breaks with almost any `react` internal change (`react-tree-walker` even has a disclaimer regarding the very same issue - https://github.com/ctrlplusb/react-tree-walker).

I've looked into fixing `react-async-bootstrapper` briefly, but wasn't able to fix it yet, it's trickier than it appears to be.

Given all of that, I'd suggest to postpone v16.6  compatibility as we won't be able to use never features anyway until react team adds support for SSR for Suspense and Concurrent mode. They've already started working on that so hopefully it won't take long - https://twitter.com/TechnocraticArt/status/1056611281317384193 .

When SSR for Suspense and Concurrent mode is supported by react itself we'll be able to use `React.lazy` for code splitting and remove `react-async-bootstrapper` altogether and hopefully react-apollo  will switch to suspense 'way' of working as well so we won't have to call `getDataFromTree` anymore.

Issue #213, opened 2018-12-05
fix setting react context when mounting async components


tutok:
**What should be changed/improved?**
react context is not set when mounting async components via react-async-bootstrapper2

**Suggested solution**
https://reactjs.org/docs/code-splitting.html#reactlazy
https://github.com/smooth-code/loadable-components
https://github.com/smooth-code/loadable-components/blob/master/packages/server/README.md

tutok:
connected with #112

PR #226, opened 2018-12-10
React 16.6 compatibility 

tadeuszwojcik:
Improvements and fixes of React v16.6 context handling when rendering server side.

**Other information:**

Closes https://github.com/deity-io/falcon/issues/112 , https://github.com/deity-io/falcon/issues/213

## Detailed changelog

- (falcon-client, shop-with-blog/client): replaced `react-async-component` and `react-async-bootstrapper2` packages with `@loadable` components (https://www.smooth-code.com/open-source/loadable-components/) - fixes issue with context handling in React 16.6 - not more react tree traversal on the server  🚀 
- (falcon-client, shop-with-blog/client, ecommerce-uikit): updated `react-apollo` to v2.3.2 that doesn't have custom react tree traversal logic on the server - uses `renderToString` - fixes issue with context handling in React 16.6 🚀 
- link prefetch/preload support added (https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c ) 🚀 

- updated `react` and `react-dom` dependencies to **v16.6.3** in all packages 🚀 💥 

- updated `react-router-dom` to `v4.4.0-beta.6` - previous version was producing warning with loadable components (https://github.com/ReactTraining/react-router/issues/6497)

- updated webpack to `v4.24.0` - previous versions weren't compatible with `@loadable` components

- fixed issue in `falcon-ui` with SSR rendering of `Details` component 🐛 

- added error boundary in `shop-with-blog\client` - `@loadable` components don't have error handler (as we had in react-async-component) 🆕 

- faster SSR - instead of 5 `renderToString` calls we have only one now (plus whatever `react-apollo` does inside) 🚀 

**Before**

current-timings
**After**
new-timings

#### All those changes make falcon compatible with 16.7.alpha as well ( **hooks** 🎉 )  

Remaining issues:
- I've found issue with `@loadable` components that was causing it's webpack plugin to hang in production mode -  https://github.com/smooth-code/loadable-components/issues/179   and currently using workaround for it (that saves loadable-stats.json to `build/public` dir not `build`) - that workaround it seems to affect hot reloading logic a little bit - to many rebuilds from time to time, also sometimes after building for production and then again for dev it seems to load incorrect version, requires restart.
IMHO it's acceptable as usual workflow works fine and hopefully fix for that will be implemented soon in `@loadable` components (or I'll create pull request for that)


examples/shop-with-blog/client/src/theme.js
```diff
@@ -206,3 +206,12 @@ export const deityGreenTheme = createTheme({
     }
   }
 });
+
+export const normalizeCss = {
```
mateuszwozniak: Let's rename it to `globalCss` so it's more clear what it's intended for.

tadeuszwojcik: Done.

UPGRADE.md
```diff
@@ -6,11 +6,39 @@ on how to migrate your previously generated Falcon-based project to a newer vers
...
 
+## Falcon 0.2.0 to 0.3.0
+
+### Falcon-Client 0.3.0
+
+If you were using `Async` components, please change them `loadable` instead:
+
+```diff
+- import AsyncComponent from 'src/components/Async';
++ import loadable from 'src/components/loadable';
```
Borales: @tadeuszwojcik could you please put a link to the file (from `master` probably) where developers could get this module from?

tadeuszwojcik: done

PR #231, opened 2018-12-13
Bundle size optimizations 

tadeuszwojcik:
**🎄 🎁 🎅 🎄  Early Christmas gift  for falcon 🚀 🎄 🎁 🎅 🎄** 

Changes in this pull request **reduce initial bundle size (vendor & client) by ~22%** (minified code bundle size) 🎉 🎉 🎉 

**Before - 666KB minified (not compressed) 👿**

bundle-size-before-changes

**After - 522 KB minified (not compressed) 😇** 

bundle-size-after-changes

### DIFF
--- a/examples/shop-with-blog/client/src/pages/shop/components/Sidebar/SidebarContainer.js
+++ b/examples/shop-with-blog/client/src/pages/shop/components/Sidebar/SidebarContainer.js
@@ -3,6 +3,15 @@ import PropTypes from 'prop-types';
 import { CloseSidebarMutation } from '@deity/falcon-ecommerce-uikit';
 import { SidebarQuery } from './SidebarQuery';
 
+// ready timeout is set based on the assumption how lighthouse measures
+// time to interactive to improve perceived performance
+
+// (TTI measures the time from Navigation Start until the page's resources are loaded
+// and the main thread is idle (for at least 5 seconds))
+// https://developers.google.com/web/updates/2018/05/first-input-delay
+
+const READY_TIMEOUT_MS = 6000;
+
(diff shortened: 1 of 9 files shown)


packages/falcon-errors/src/errors.js
```diff
@@ -1,9 +1,30 @@
-const { ApolloError, AuthenticationError } = require('apollo-server-errors');
```
Borales: @tadeuszwojcik please remove `apollo-server-errors` dependency in `packages/falcon-errors/package.json`

tadeuszwojcik: Done.

examples/shop-with-blog/client/src/pages/shop/components/Sidebar/SidebarContainer.js
```diff
@@ -17,24 +26,40 @@ export class SidebarContainer extends React.Component {
   }
 
   componentDidMount() {
-    // wait with loading sidebar contents until browser is idle
-    // as it's not high prority task and would delay time to interactive
     const setReady = () => {
```
tutok: _I am aware that this note applies to the code which was not written under this PR_

I am just thinking, maybe if it is worth to extract "content deferred loading" functionality into the separate component? lets say: 

```
<WindowReady force={true|false}>
 // content which should be loaded when timeout
</WindowReady>
```
flag `force` could be used to load earlier, before the timeout, as some external condition require it. In our case when the sidebar is opened by the user.

My motivation to do this is a rather high probability to load some other content in a similar manner (let say lightboxes)

Or it does not make sense as we can do this with React.lazy?

PR #235, opened 2018-12-14
update emotion version to v10.05 & workaround for not rendering styles on the server after SSR error

tadeuszwojcik:


tadeuszwojcik:
I wasn't able to track exact cause why styles aren't rendered on the server after SSR error, my suspicion is something introduced in the logic of `react-apollo` in https://github.com/apollographql/react-apollo/pull/2533  in `getDataFromTree`, maybe something isn't cleaned up properly when error occurs during rendering of react inside `getDataFromTree`, not sure. 

Outcome of that behavior is that emotion's cache provider is not discarder/but still used on subsequent server renders.

Before error occurs this line https://github.com/emotion-js/emotion/blob/master/packages/core/src/context.js#L54 always evaluates to `true` so new cache provider is created on each react render, after error on the server it's not the case anymore and `context ` in that line is already set, really strange, and doesn't happen if I remove call to `getDataFromTree` and trigger error in different way, hence my suspicion with `react-apollo`, but there is a slight chance that it's something else as it's really hard to track down what exactly is going on there.
 Last commit in this pull request implements workaround I barely found, seems to be working, but please test it. 
More long term approach would be nice, like creating more basic reproducible example so folks from apollo team could look into that.

PR #247, opened 2018-12-25
Payments

### LINKED ISSUE #94: GraphQL - JSON (any) structure
There are cases we when have to get unknown data structures:
- getting config values with optional keys (client)
- getting unknown data structures for a custom data (server)

Introduce `JSON` type (for server-side requests) and `@json` or `@any` (for client-side requests) directive to handle unknown data structures

### LINKED ISSUE #233: Payments
Introduce support for external Payment systems

- [ ] support multi-api payments
- [ ] support callback hooks (endpoints)
- [ ] tests and docs

Borales:

**Other information:**
Closes #94 and resolves #233


examples/shop-with-blog/client/src/pages/shop/components/CreditCard.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import CreditCardInput from 'react-credit-card-input';
+import { Box, Input, Label } from '@deity/falcon-ui';
+
+class CreditCard extends React.Component {
```
tutok: Personally, I would use `formik` to get CC data. One more time. I would get state management, values validation, label/placeholder translations out of the box. So this component could be simplified and any manual checks will not be required.

Looks like somethink like this should be possible:

```
import {  FormField } from '@deity/falcon-ecommerce-uikit';

<CreditCardInput
            inputComponent={FormField}
            cardNumberInputProps={{ name: 'number' }}
            cardExpiryInputProps={{ name: 'expiry' }}
            cardCVCInputProps={{ name: 'cvc' }}
          />
```

packages/falcon-client/src/bootstrap/endpoints.js
```diff
@@ -0,0 +1,75 @@
...
+            }
+          })
+        );
+
+        const route = router.route(endpoint);
+        const routeIndex = router.routes().router.stack.indexOf(route);
+        // Rearranging proxied routes - we have to make sure
+        // these routes are added at the beginning of the router stack
+        router.routes().router.stack.splice(routeIndex, 1);
+        router.routes().router.stack.unshift(route);
```
tutok: I think here is too many assumptions about route order. Please register this `endpoints` configuration on `koa-router` `router` in proper order, and remove this code.

packages/falcon-client/src/bootstrap/endpoints.js
```diff
@@ -0,0 +1,75 @@
...
+ * @param {string} redirs.cancel Cancel page
+ * @param {boolean} isDebug Debug flag
+ */
+export const endpoints = (router, serverUrl, redirs, isDebug = process.env.NODE_ENV !== 'production') => {
+  if (!serverUrl) {
+    Logger.warn('"serverUrl" must be passed in your "bootstrap.js" file.');
+    return;
+  }
+  const endpointsConfig = url.resolve(serverUrl, '/endpoints');
+  fetch(endpointsConfig)
```
tutok: I think I would not fetch `endpointsConfig` here, I will provide it as a `param` to `endpoints` function. 

the main reason are unit tests. It will be much easier to mock that configuration,

packages/falcon-client/src/bootstrap/endpoints.js
```diff
@@ -0,0 +1,75 @@
...
+ * @param {string} serverUrl Falcon-Server URL
+ * @param {object} redirs Map of redirections
+ * @param {string} redirs.success Success page
+ * @param {string} redirs.failure Failure page
+ * @param {string} redirs.cancel Cancel page
+ * @param {boolean} isDebug Debug flag
+ */
+export const endpoints = (router, serverUrl, redirs, isDebug = process.env.NODE_ENV !== 'production') => {
+  if (!serverUrl) {
+    Logger.warn('"serverUrl" must be passed in your "bootstrap.js" file.');
```
tutok: I think above warn message should be: `'"serverUrl" is required.'`. I believe we should not mention about `bootstrap.js` file. I would also change `warn` into `error`

I think that `router` param also is required.

examples/shop-with-blog/client/package.json
```diff
@@ -18,24 +18,30 @@
   },
   "dependencies": {
     "@deity/falcon-client": "^0.3.0",
+    "@deity/falcon-errors": "^0.1.0",
+    "@deity/falcon-payment-plugin": "^0.0.1",
+    "@deity/falcon-adyen-plugin": "^0.0.1",
+    "@deity/falcon-paypal-plugin": "^0.0.1",
```
tutok: I am not convinced that we need to create a separate package for any kind of payment method. I know that we have problems with code organization and component replacement but adding more packages doesn't solve done of them

packages/payment/falcon-adyen-plugin/src/AdyenCCPlugin.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  render() {
+    const CreditCardInput = this.props.creditCardInput;
+    const CreditCardInputOutput = (
+      // @ts-ignore
+      <CreditCardInput onCompletion={(creditCardData: CreditCardData) => this.encryptCreditCard(creditCardData)} />
+    );
+
+    return this.props.template({
```
tutok: Instead of rendering `template` I would rather expose necessary functionality via render props

packages/payment/falcon-adyen-plugin/src/AdyenCCPlugin.tsx
```diff
@@ -0,0 +1,68 @@
...
+export interface CreditCardData {
+  number: string;
+  expiry: string;
+  cvc: string;
+  name: string;
+}
+
+export class AdyenCCPlugin extends PluginModel<AdyenProps> {
+  static icon: string =
+    'https://raw.githubusercontent.com/Adyen/adyen-magento2/develop/view/base/web/images/logos/creditcard.png';
```
tutok: should not we have a local copy of that image?
why we can not use our [template mechanism](https://github.com/deity-io/falcon/blob/payments/examples/shop-with-blog/client/src/theme.js#L31) to provide that icon?

examples/shop-with-blog/client/bootstrap.js
```diff
@@ -1,8 +1,15 @@
...
+const { endpoints } = require('@deity/falcon-client/src/bootstrap/endpoints');
 
 export default {
-  config: { ...config }
+  config: { ...config },
   // onServerCreated: server => { console.log('created'); },
   // onServerInitialized: server => { console.log('initialized'); },
   // onServerStarted: server => { console.log('started'); }
+  onRouterCreated: router =>
+    endpoints(router, config.apolloClient.httpLink.uri, {
```
tutok: `endpoints` is not the best name here, as it suggests that generic configuration could be provided while it is not. It accepts only a payment config. We should change the name or change the way of specifying the redirect config to be more generic. I would choose the second option.

examples/shop-with-blog/client/src/pages/shop/Checkout/PaymentSelector.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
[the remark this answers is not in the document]
Borales: @tutok why? we have `.js` files in our example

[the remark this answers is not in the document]
tutok: ahhh

examples/shop-with-blog/client/src/pages/shop/Checkout/PaymentSelector.js
```diff
@@ -0,0 +1,72 @@
...
+          }}
+          methods={availableMethods}
+          paymentsConfig={paymentsConfig}
+        >
+          {({ code, title, onSelect, icon, pluginComponent }) => (
+            <FlexLayout key={code} my="xs" css={{ alignItems: 'center' }}>
+              <Radio size="sm" name="payment" id={`opt-${code}`} value={code} onChange={onSelect} />
+              <Label mx="sm" flex="1" htmlFor={`opt-${code}`}>
+                {icon && <Image src={icon} mr="xs" mb="xs" css={{ verticalAlign: 'middle' }} />}
+                {title}
```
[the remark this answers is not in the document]
Borales: @tutok comes from the backend

examples/shop-with-blog/client/src/pages/shop/CheckoutFailure.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { H1, Text, FlexLayout } from '@deity/falcon-ui';
+
+const CheckoutFailure = () => (
+  <FlexLayout my="xxl" flexDirection="column" alignItems="center">
+    <H1 mb="xl">Order failed</H1>
+    <Text>Unfortunately we could not process your payment, try again later!</Text>
```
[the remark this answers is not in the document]
Borales: @tutok we definitely should, this will be done in a separate task https://github.com/deity-io/falcon/issues/344

packages/falcon-client/src/bootstrap/endpoints.js
Borales: @tutok `'"serverUrl" is required.'` message gives no context about the nature of this error at all, therefore we need to be explicit in this case.

examples/shop-with-blog/client/src/pages/shop/components/CreditCard.js
```diff
@@ -0,0 +1,77 @@
...
+  state = {
```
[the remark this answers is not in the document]
Borales: @tutok I will, but seems like we don't have a single style on how to define initial state

Screenshot 2019-03-13 at 03 08 59

packages/falcon-client/src/bootstrap/endpoints.js
tutok: I see that method `endpoints`  have the following contract:
`router` - required
`serverUrl` - requred
`redirs` - requred
`isDebug` - optional
so, I still do not understand how we can log `warn` when the contract is not met?

If you think that lack of `serwerUrl` is correct and you would like to only warn the user, then you need to move the `if (!serverUrl) { ....` check outside of  `endpoints` method. As technically method `endpoints` can not do its job without `serwerUrl` value.

packages/falcon-proxy-endpoints/package.json
```diff
@@ -0,0 +1,38 @@
+{
```
mateuszwozniak: I'd suggest to remove this package and move the implementation to `falcon-server-env.

packages/payment/falcon-adyen-plugin/src/AdyenCCPlugin.tsx
mateuszwozniak: Yes, I think it would be a better option. API will be almost the same and it will be more react-way.

packages/falcon-payment-plugin/src/PlainPayment.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { PluginModel } from './components/Plugin';
+
+export class PlainPayment extends PluginModel {
```
mateuszwozniak: What about `SimplePayment`? Seems to be a bit better name for me.

examples/shop-with-blog/client/package.json
mateuszwozniak: Initially I thought that separate modules are good for those payments, but it seems to me that `@deity/falcon-adyen-plugin` and `@deity/falcon-paypal-plugin` could go to `@deity/falcon-ecommerce-uikit`. `@deity/falcon-payment-plugin` can stay as it is. 
Main reason behind that decision is that we have couple of lines of code in these 2 modules and dozen of lines of configs (package.json, tests linters etc). So it seems to be that it'a a bit too much noise to signal ratio :)

examples/shop-with-blog/client/src/pages/shop/Checkout/PaymentSelector.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+PaymentSelector.propTypes = {
+  availableMethods: PropTypes.arrayOf(PropTypes.shape({})),
+  onPaymentSelected: PropTypes.func,
+  paymentsConfigQuery: PropTypes.shape({})
+};
+
+export default graphql(GET_CONFIG, {
```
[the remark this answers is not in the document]
mateuszwozniak: Yes, `<Query/>` could be used here. `graphql()` should rather be used in the places where we need to pass data retrieved from GraphQL as props to the component (when it's needed outside of `render()`) method.

examples/shop-with-blog/client/src/pages/shop/components/CreditCard.js
mateuszwozniak: Yes, I think that would simplify the code a bit.

packages/falcon-payment-plugin/src/PlainPayment.tsx
tutok: `BankTransfer` ?

examples/shop-with-blog/client/src/pages/shop/Checkout/PaymentSelector.js
```diff
@@ -0,0 +1,72 @@
...
+    return (
+      <Box my="md">
+        <Payment
+          creditCardInput={CreditCard}
+          template={PluginTemplate}
+          onPaymentSelected={onPaymentSelected}
+          plugins={{
+            adyen_cc: AdyenCCPlugin,
+            paypal_express: PayPalExpressPlugin,
+            checkmo: PlainPayment
```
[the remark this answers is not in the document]
Borales: @tutok such information is going to be sent to the customer's email (depending on the backend settings)

packages/falcon-client/src/bootstrap/endpoints.js
Borales: changed

examples/shop-with-blog/client/package.json
Borales: as discussed - keeping them in separate packages

packages/payment/falcon-adyen-plugin/src/AdyenCCPlugin.tsx
Borales: @tutok in our case - that's not an SVG icon, so just for the simplicity of this solution and keeping the bundle file without drastic changes  - we use image URL. 
I checked a few rollup plugins to encode the image, but none of them could properly handle the image file. Since we keep the template/render mechanism within the example - anyone could override `renderIcon` method in `examples/shop-with-blog/client/src/pages/shop/Checkout/components/PaymentMethodItem.js` with their own icon (otherwise - the default one will be used).

packages/falcon-client/src/bootstrap/endpoints.js
Borales: not really sure I got your point.

To avoid double-configuring on FC and FS sides - we need to fetch this data on the app start-up from Falcon-Server. This function acts as an extra hook to the Falcon-Client app (although, after last changes - it has been split to 2 different functions)

examples/shop-with-blog/client/src/pages/shop/components/CreditCard.js
Borales: @tutok @mateuszwozniak the number of checks done in this component was caused by CreditCardInput package itself (initially it was chosen to support our input classes and easy styling). It does all the checks and validations (invalid date/cvc/credit number etc) and provides required translations, **but** it does not provide any "completed" event (https://github.com/medipass/react-credit-card-input/issues/39) so any external code cannot be notified on successful fulfillment.

examples/shop-with-blog/client/bootstrap.js
```diff
@@ -8,11 +8,12 @@ const redirects = {
...
 };
+const serverUrl = config.graphqlUrl || config.apolloClient.httpLink.uri;
 
 export default {
   config: { ...config },
   // onServerCreated: server => { console.log('created'); },
   // onServerInitialized: server => { console.log('initialized'); },
   // onServerStarted: server => { console.log('started'); }
-  onRouterCreated: async router => bootstrap(router, config.apolloClient.httpLink.uri, redirects)
+  onRouterCreated: async router => bootstrap(router, serverUrl, redirects)
```
tutok: personally, I would not call this method `bootstrap`, it is too vague, I would suggest something like `configureProxy`

Borales: @tutok I don't mind changing this name, but `configureProxy` in this case is too narrow and specific (as it could perform not only proxy-endpoints in the future). This is more like a "runtime-bootstrap" hook (fetching some basic config from the Falcon-Server).

@mateuszwozniak any suggestions on how to name it?

mateuszwozniak: `configureServer`, `injectConfiguration`?

Borales: as it relates to the server-side of Falcon-Client app - I agree with `configureServer`, 
what would you say @tutok ?

Borales: updated

packages/falcon-magento2-api/src/index.test.js
```diff
@@ -21,7 +21,13 @@ const apiConfig = {
   eventEmitter: ee,
   gqlServerConfig: {
     schema: makeExecutableSchema({
-      typeDefs: [BaseSchema, Schema]
+      typeDefs: [BaseSchema, Schema],
+      resolvers: {
+        PlaceOrderResult: {
+          // empty resolver
```
mateuszwozniak: I can clearly see it's an empty resolver... is there a better explanation that could be placed here?

Borales: @mateuszwozniak I updated this comment. Even though the warning shows how to disable it - it doesn't do trick.

examples/shop-with-blog/client/src/pages/shop/Checkout/components/CreditCard.js
```diff
@@ -0,0 +1,123 @@
...
+              customTextLabels={{
+                expiryError: {
+                  invalidExpiryDate: t('creditCard.errors.invalidExpiryDate'),
+                  monthOutOfRange: t('creditCard.errors.monthOutOfRange'),
+                  yearOutOfRange: t('creditCard.errors.yearOutOfRange'),
+                  dateOutOfRange: t('creditCard.errors.dateOutOfRange')
+                },
+                invalidCardNumber: t('creditCard.errors.invalidCardNumber'),
+                invalidCvc: t('creditCard.errors.invalidCvc'),
+                cardNumberPlaceholder: t('creditCard.number'),
```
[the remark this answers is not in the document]
Borales: @tutok updated

PR #259, opened 2019-01-10
scss and css modules support in falcon-client

### LINKED ISSUE #215: Add sass/scss plugin for falcon-client
In theory, all razzle plugins should be compatible with falcon-client. This means that npm package `razzle-plugin-scss` should work. But is not.

because the `razzle-plugin-scss` is not properly published on npm we need to install `@davidqhr/razzle-plugin-scss`, moreover, its implementation uses razzle internals while it should not. To work around this we need to install `razzle` too.

I believe It should be simpler than that, so we should prepare own version of sass plugin for falcon-client.  

Here is a temporary solution to have sass/scss support:
- navigate to dir were `your-app/client` is located 
- `yarn install @davidqhr/razzle-plugin-scss` 
- `yarn install razzle` (because the above plugin uses razzle internals while it should not!)
- open `falcon-client.build.config` and add:
  ```
    const falconClientPaths = require('@deity/falcon-client/build-utils/paths');

    module.exports = {
      plugins: [
        {
           func: require('@davidqhr/razzle-plugin-scss'),
          options: {
            sass: {
              dev: {
                sourceMap: true,
                includePaths: [falconClientPaths.appSrc]
              },
              prod: {
                 sourceMap: false,
                includePaths: [falconClientPaths.appSrc]
              }
            }
          }
        }
      ]
    };
  ``` 
after that you should be able to use sass. 

Here is example which shows how to use sass into your app: https://github.com/jaredpalmer/razzle/tree/master/examples/with-scss

Assuming that yo are in `my-app/client/src/` directory, add `App.scss` file with your styles, later open `App.js` and add `import './App.scss'`

tutok:
resolve #215


cane-lagerwaard:
https://github.com/deity-io/falcon-docs/issues/19

packages/falcon-client/build-utils/webpack/config/runPlugin.js
```diff
@@ -1,28 +1,16 @@
-function runPlugin(plugin, config, { target, dev }, webpack) {
-  if (typeof plugin === 'string') {
-    // Apply the plugin with default options if passing only a string
-    return runPlugin({ name: plugin }, config, { target, dev }, webpack);
-  }
+// / { target, dev }
```
Borales: @tutok please remove this line

tutok: done

tutok:
@mateuszwozniak docs will be added under [falcon-docs #19](https://github.com/deity-io/falcon-docs/issues/19)  task

PR #280, opened 2019-01-24
Dynamic menu

### LINKED ISSUE #275: Dynamic main menu
**Is your feature request related to a problem? Please describe.**
Menu rendered in Navbar is taken from the hardcoded configuration file, which causes that it needs to be synchronized with Magento manually. 

**Describe the solution you'd like**
add graphQL query resolver for `/rest/default/V1/menu` Magento endpoint and use that Query to make Navbar content dynamic.

tutok:
resolve #275


packages/falcon-magento2-api/src/index.js
```diff
@@ -65,6 +66,31 @@ module.exports = class Magento2Api extends Magento2ApiBase {
...
+    const { data } = this.convertKeys(response);
+
+    const mapMenu = x => {
+      if (Array.isArray(x)) {
+        return x.length > 0 ? x.map(mapMenu) : [];
+      }
+
+      return {
+        ...x,
+        urlPath: urlJoin(x.urlPath, undefined, { leadingSlash: true }),
```
Borales: @tutok we could simply add a `/` prefix depending on its value (without using a dedicated package).

```suggestion
        urlPath: x.urlPath.startsWith('/') ? x.urlPath : `/${x.urlPath}`,
```

tutok: and what in a situation when we will have urlPath like `//foo/bar`?

Borales: @tutok if `/` is already there - we won't be adding it. If there are 2 `/` there - we shouldn't be fixing it (maybe there's such intention behind it)

tutok: I decided to use external package because I feel that I can not assume what kind of url exacly Magento can return. While we need to normalize these to be able to use them in falcon-client. So instead of writeing code myself to recognize all use cases
- if leadingSlash exists
- if protocol exists
- is trailingSlash exists
found that `proper-url-join` cover all of them 

```
urlJoin('https://www.google.com/', undefined, { leadingSlash: true })
> "https://www.google.com"

urlJoin('www.google.com/', undefined, { leadingSlash: true })
> "/www.google.com"

urlJoin('//www.google.com/', undefined, { leadingSlash: true })
> "/www.google.com"

urlJoin('www.google.com//', undefined, { leadingSlash: true })
> "/www.google.com"

urlJoin('www.google.com', undefined, { leadingSlash: true })
> "/www.google.com"

```

PR #299, opened 2019-02-03
Suggestion on how to re-use apollo-link-http mock

Borales:


packages/falcon-client/src/server.test.js
```diff
@@ -82,20 +79,26 @@ describe('Server', () => {
...
-                  backendConfig: () => ({
-                    locales: ['en-US'],
-                    activeLocale: 'en-US'
-                  })
-                }
-              }
-            })
-          })
+        httpLink: {
+          typeDefs: `
```
tutok: I am not sure about this, I will leave 
```
  { 
    typeDefs: [BaseSchema]
  }
```

where `BaseSchema` is imported from `@deity/falcon-server`
I think it should be technically possible, am I right?

Borales: @tutok I just wanted to get rid of `falcon-server` dependency, as it looks a bit weird to have it in `falcon-client` package. At this moment - we simply need 1 Query for `falcon-client/server` tests. 

On the other hand - we test our packages within a monorepo setup, so all these packages will be installed anyways, so you can keep falcon-server, yes. And it is possible to import typeDefs from FS and pass them as above.

PR #306, opened 2019-02-05
Add browserslist support for js

### LINKED ISSUE #293: IE11 bind bug
**Describe the bug**
In IE11, a a JS error breaks Falcon.

**To Reproduce**
1. Run the server in production mode
2. Build the client and run in production mode
3. Observe broken JS functionality in IE11

**Expected behavior**
In IE11, functionality should be comparable to other browsers.

```
SCRIPT5007: Unable to get property 'bind' of undefined or null reference
vendors.1796c3c3.chunk.js (9,169993)
```

 - Windows 10
 - IE11

tutok:
this PR adds full support for `.browserslistrc`, which means all necessary javascript polyfills will be added based on its configuration. 

fix #293


tutok:
I need to do one more test of stand alone project generated via CFA (outside of our monorepo) to validate correctness of solution. I think that polyfills will be added properly but it is better to chcek it.

tutok:
I found confirmation of my assumptions that there is some issue in `babel`/`webpack` pipeline https://github.com/babel/babel-loader/issues/567.

I confirmed earlier today that proposed solutions works, but actually these solutions have a huge impact on our entire webpack configuration, especially for `development`, as we have `HMR`.

PR #309, opened 2019-02-11
Filters - first stage

mateuszwozniak:
This pull request provides first stage of implementation of filters.

New feature - filters. Implementation covers:
- required changes in schema
- changes in falcon-magento2-api class to handle filters correctly
- changes in falcon-ecommerce-uikit that provide implementation of search logic (storing filters in url, restoring them, reacting for history change, refreshing search results on filters change etc).

Detailed documentation will be provided in near future as second stage of the work related to filters.


examples/shop-with-blog/client/src/pages/shop/Category.js
```diff
@@ -1,30 +1,34 @@
...
-            activeSortOrder={activeSortOrder}
-            setSortOrder={setSortOrder}
-          />
-        )}
-      </CategoryProductsQuery>
-    )}
-  </SortOrdersProvider>
+  <SearchContext.Consumer>
+    {({ state, availableSortOrders, setSortOrder }) => {
+      const activeSortOrder = state.sort || availableSortOrders[0];
```
Borales: @mateuszwozniak shouldn't `activeSortOrder` be set in the `SearchContext.Consumer`?

packages/falcon-ecommerce-uikit/src/Search/types.d.ts
```diff
@@ -0,0 +1,43 @@
+export enum SortOrderDirection {
+  asc = 'asc',
+  desc = 'desc'
+}
+
+export enum FilterOperator {
+  eq = 'eq',
```
Borales: @mateuszwozniak so we're adding enums? 😄

packages/falcon-magento2-api/src/index.js
```diff
@@ -110,6 +110,16 @@ module.exports = class Magento2Api extends Magento2ApiBase {
    */
   async categoryProducts(obj, params) {
     const query = this.createSearchParams(params);
+
+    this.addSearchFilter(params, 'visibility', '2,4', 'in');
```
Borales: @mateuszwozniak please add a note for this line too

examples/shop-with-blog/client/src/App.js
```diff
@@ -65,37 +66,41 @@ const App = ({ online }) => (
...
-                    </Sidebar>
-                  )}
-                </SidebarContainer>
-              </ErrorBoundary>
-            </AppLayout>
-          </ThemeProvider>
-          {ThemeEditorComponent && <ThemeEditorComponent {...props} side="left" />}
-        </React.Fragment>
+        <SearchProvider>
+          {() => (
```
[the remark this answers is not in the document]
mateuszwozniak: Good point. That's a leftover from first version of the api :) I'll fix that in a minute. When it comes to the "root" components I think that it's good we have those here so people see what exactly builds up the main application. What do you think?

packages/falcon-magento2-api/src/index.js
mateuszwozniak: 👍

packages/falcon-ecommerce-uikit/src/Search/types.d.ts
mateuszwozniak: What's the downside of that? :)

examples/shop-with-blog/client/src/App.js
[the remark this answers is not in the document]
Borales: @mateuszwozniak the last one - it was just a question to try to get rid of so many nested levels.
They can stay here without any changes. 👍

packages/falcon-ecommerce-uikit/src/Search/types.d.ts
Borales: @mateuszwozniak the possible usage of eject-ts

examples/shop-with-blog/client/src/pages/shop/Category.js
mateuszwozniak: Yes I moved it to `SearchProvider`.

examples/shop-with-blog/client/src/App.js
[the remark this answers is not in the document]
mateuszwozniak: done

packages/falcon-shop-extension/src/schema.graphql
```diff
@@ -100,9 +100,56 @@ input ShopPageQuery {
...
+  """
+  neq
+
+  """
+  less then
+  """
+  lt
+
+  """
+  less then or equals
```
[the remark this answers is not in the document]
mateuszwozniak: done

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+  };
+
+  render() {
+    return (
+      <SearchContext.Provider
+        value={{
+          state: {
+            ...this.state,
+            // if there's no sort set yet then return first available option (it's considered as default one)
+            sort: this.state.sort || this.props.sortOrders[0]
```
Borales: @mateuszwozniak we cannot consider `name: 'Price ascending',` as our default sort order, the backend can have its own default sorting order (usually "created date" or something), so setting first available sort order is not right (correct me if I'm wrong)

packages/falcon-magento2-api/src/index.js
mateuszwozniak: done

Borales: @mateuszwozniak it was `2,4` for a reason... but I don't remember which one :) ok, resolving this one

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
mateuszwozniak: We have to display something in the UI (currently set value) even if user didn't select anything by himself. These sorting orders in the final version should come from the backend (that's why we keep these in apollo state) so it's up to backend to return the default option as the first one.

Borales: @mateuszwozniak so it will be available after we switch to the backend "sort orders" completely, right? currently, we still have this hardcoded. Looks like Magento has "position" as a default sort order.

Borales: @mateuszwozniak if that's so - please resolve this comment by yourselve, the PR itself is approved!

mateuszwozniak: Yes, exactly. We'll get exact values when we connect backend endpoint for that (which we don't have yet).

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    const filters = this.state.filters.filter(filter => filter.field !== field);
+
+    this.updateURL({
+      ...this.state,
+      filters
+    });
+  };
+
+  private historyUnlisten: Function;
```
tutok: maybe we can merge this declaration with default value initialization from [ctor](https://github.com/deity-io/falcon/pull/309/files#diff-269cc7cda5d1b5b4e58c32320972d090R32)?

```suggestion
  private historyUnlisten = () => {};
```

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  updateURL(state: SearchState) {
+    const queryString = this.props.searchStateToURL!(state);
+    this.props.history.push(`${this.props.location.pathname}?${queryString}`);
+  }
+
+  restoreStateFromURL = (location: any) => {
```
tutok: we can take `location` type from `history` which is used by `react-router` anyway
```
// eslint-disable-next-line
import { Location } from 'history';
```

(maybe we should consider to add `history` into dependencies.)

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
tutok: to be honest I think UI should not be aware of a default order. I believe that "generic data provider" can return data not ordered (e.q. it is not necessary to specify any sort order when querying SQL database `SELECT * FROM foo`). The same situation is with `filters`.

If I, as a consumer of this API, would like to have defaults I would like to specify them:
```jsx
  <SearchProvider filter={/*default*/} filters={/*available*/} sortOrder={/*default*/} sortOrders={/*available*/}> 
    { 
       //...
    }
  </SearchProvider >
```

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+const SearchProviderWithSortOrders = graphql<any, { sortOrders: SortOrder[] }>(SORT_ORDERS_QUERY, {
+  // remap data received from apollo - return sortOrders directly
+  props: ({ data, ownProps }) => ({
+    sortOrders: data!.sortOrders,
+    ...ownProps
+  })
+})(SearchProviderImpl);
+
+// wrap everything in router so SearchProviderImpl has access to history and location
+export const SearchProvider = withRouter(SearchProviderWithSortOrders);
```
tutok: we have `falcon-ecommerce-uikit/Router` component which provide all `withRouter` props as render props, so then we can move it to `render` function of `SearchProviderImpl` component and improve readability

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+        }}
+      >
+        {this.props.children}
+      </SearchContext.Provider>
+    );
+  }
+}
+
+// wrap SearchProviderImpl with SORT_ORDERS_QUERY so sort orders are passed as props to SearchProviderImpl
+const SearchProviderWithSortOrders = graphql<any, { sortOrders: SortOrder[] }>(SORT_ORDERS_QUERY, {
```
tutok: I am just curious why we can not use our `Query` component?

tutok:
in general looks good! 👍

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
mateuszwozniak: It works kind of that way. It's not a default, but it's just first on the list, because we have to display something (dropdown for sort order cannot be empty).

packages/falcon-ecommerce-uikit/src/Search/searchStateFromURL.ts
```diff
@@ -0,0 +1,41 @@
+import qs from 'qs';
```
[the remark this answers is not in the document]
mateuszwozniak: I was thinking about that but the problem is that we need custom features that `URLSearchParams` doesn't have - serializing/deserializing of nested objects, custom encoding etc.

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+export const SORT_ORDERS_QUERY = gql`
+  query SortOrdersQuery {
+    sortOrders @client
+  }
+`;
+
+interface SearchProviderProps extends RouteComponentProps {
+  searchStateToURL?(state: SearchState): string;
+  searchStateFromURL?(url: string): SearchState;
```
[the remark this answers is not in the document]
mateuszwozniak: User doesn't have to pass that (so it's optional) but there's a default value (in defaultProps) in case user doesn't provide it.

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
mateuszwozniak: That was my first approach because I wanted to avoid so many HOC's but we cannot move it to the `SearchProviderImpl.render()` because I need to have access to things provided by the router via `this.props` (e.g. in `SearchProviderImpl.updateURL()` method).

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
mateuszwozniak: Mostly because I wanted to keep things consistent. We have `withRouter` and `graphql` here so I wanted to use both in the same way - to avoid rendering component inside component and then that component again inside another component. The code is just shorter with the current approach.

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+        delete state.sort;
+      }
+    }
+
+    return state;
+  }
+
+  getFullSortOrderDefinition(sort: SortOrder) {
+    return this.props.sortOrders.find(
+      (item: SortOrder) => item.field === sort.field && item.direction === sort.direction
```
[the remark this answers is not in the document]
mateuszwozniak: Good point - changed.

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
```diff
@@ -0,0 +1,172 @@
...
+    );
+  }
+
+  setFilter = (field: string, value: string[], operator = 'eq' as FilterOperator) => {
```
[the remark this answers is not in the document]
mateuszwozniak: True, good catch - changed.

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
mateuszwozniak: True, done :)

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.tsx
mateuszwozniak: Yes, I changed that. I didn't add the dependency though as I'll be adding things to that file soon, so I'll revisit that history related stuff.

Issue #325, opened 2019-02-24
Homescreen application does not show icon set in the manifest


cane-lagerwaard:
**Describe the bug**
When you add the shop-with-blog example to homescreen, it does not take the icon set in the web manifest

**To Reproduce**
1. Go to `snapshot.deity.io`
2. Add the website to homescreen
3. Notice the app's icon is a screenshot from the website, instead of the icon set in manifest

**Expected behavior**
The app should take the icon's set in the manifest

**Smartphone (please complete the following information):**
 - Device: iPhone 6s
 - OS: iOS 12
 - Browser: Safari

**Additional info**
It appears that `svg` is not supported and `png` should be used instead - this is also the case on chromium browsers: https://bugs.chromium.org/p/chromium/issues/detail?id=578122

tutok:
when https://app.zenhub.com/workspaces/falcon-overview-5bb732c64b5806bc2beb8408/issues/deity-io/falcon/306 will be merged please add to `polyfill` chunk  `pwacompact` dependencty

PR #329, opened 2019-03-04
Fix manifest

tutok:
fix: #325


tutok:
when #306 will be merged then updated `polyfill` chunk - put there `pwacompat` package

PR #330, opened 2019-03-07
Autocomplete

### LINKED ISSUE #288: Required "country" field can be skipped during the checkout
In the shop-with-blog example, when you go to checkout and fill in all the required fields except "country", you're still able to continue, but you get stuck on the next part

tutok:
resolve #288


tutok:
@Borales can you please share more context? which browser, its version, OS?
Can you please also validate it this [Select](https://falcon-ui.docs.deity.io/components-select) component also look so ugly

Borales:
@tutok sorry, my bad. It was the latest Chrome on MacOS. It probably takes the default style from the OS, but on this form - "country" looks so tiny. 

`Select` in chrome:

Screenshot 2019-03-11 at 16 43 57

The image below is from Firefox - looks slightly better, but still:

Screenshot 2019-03-11 at 16 42 01

tutok:
above issue is solved in ie10+, firefox and chrome (windows and macos)

PR #333, opened 2019-03-07
Apollo update

mateuszwozniak:
Update of ApolloClient to v2.5.1 (apollo-client + react-apollo). I've also enabled type names on client data so now caching should work correctly.


packages/falcon-client/src/service/ApolloClient.js
```diff
@@ -35,11 +34,12 @@ export function ApolloClient(config = {}) {
...
     cache
   } = config;
+
   const { httpLink, connectToDevTools, ...restApolloClientConfig } = apolloClientConfig;
-  const addTypename = false; // disabling 'addTypename' option to avoid manual setting "__typename" field
 
-  const inMemoryCache = cache || new InMemoryCache({ addTypename }).restore(initialState);
-  const apolloClientStateLink = withClientState({ cache: inMemoryCache, ...clientState });
+  const inMemoryCache = cache || new InMemoryCache().restore(initialState);
+  inMemoryCache.writeData({ data: clientState.data });
```
Borales: @mateuszwozniak I would add `client.onResetStore(() => cache.writeData({ data }));` too (just in case if somebody will call `client.resetStore()` (not just us).

packages/falcon-client/src/service/ApolloClientWeb.js
```diff
@@ -3,7 +3,7 @@ import { InMemoryCache } from 'apollo-cache-inmemory';
 import { ApolloClient } from './';
 
 export const apolloClientWeb = ({ initialState, clientApolloSchema, apolloClientConfig }) => {
-  const cache = new InMemoryCache({ addTypename: false });
```
tutok: I remember the time when we actually turned it off, to not provide types everywhere. is this decision already out of date?

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.test.tsx
```diff
@@ -41,9 +33,18 @@ describe('SearchProvider', () => {
...
+        ROOT_QUERY: {
+          sortOrders
+        }
+      })
+    });
+
     wrapper = mount(
       <MemoryRouter initialEntries={['/']}>
-        <MockedProvider mocks={dataMocks} addTypename={false}>
+        <ApolloProvider client={client}>
```
tutok: I found that you replace `MockedProvider` via `ApolloProvider`, so should not we be consistent and do same update also [here](https://github.com/deity-io/falcon/blob/apollo-update/packages/falcon-client/test-utils/FalconClient.js#L3)?

packages/falcon-client/src/service/ApolloClientWeb.js
mateuszwozniak: Yes, we wanted to enable it once we get updated ApolloClient on the front-end, so I added it to the update.

packages/falcon-ecommerce-uikit/src/Search/SearchProvider.test.tsx
mateuszwozniak: In that particular case `MockedProvider` didn't pass the data to the render prop function. I figured out that there are weird issues with `MockedProvider` outlined [here](https://github.com/apollographql/react-apollo/issues/1711) and [here](https://github.com/apollographql/react-apollo/issues/674).

That problem doesn't affect the file you mentioned because in that case `MockedProvider` just injects instance of ApolloClient and doesn't pass any data (it just prepares whole stack for testing), so I didn't want to make code more complex (as it would require at least a small config for ApolloClient instance etc).

packages/falcon-client/src/service/ApolloClient.js
mateuszwozniak: Done

packages/falcon-client/src/service/ApolloClientWeb.js
Borales: @tutok another thing is that in the upcoming Payments - there will be a `union` return type, so Apollo requires this `__typename` data to resolve it properly.

PR #337, opened 2019-03-11
Handling client cookies

### LINKED ISSUE #252: Cross-domain session cookie (server and client)
Currently, session-cookie is being passed properly between domain (client) and its subdomain (server) only.

We need to make sure it works between 2 totally different domains.

We need to handle koa session-cookie "domain" part properly and set the cookie to the origin domain.

Borales:
This PR is attempt to fix #252 and **unblock** #233 issues

Notes:
- In order to solve the "cookie" issue - we have to add another proxy "jump" between Falcon-Client and Falcon-Server apps (meaning - "no cross-domain requests on the client side")
- We don't change the cookie's `httpOnly` flag - it remains to be secure
- Due to the additional `/graphql` endpoint in Falcon-Client and proxying these requests to Falcon-Server - it could potentially reduce the performance but the "proxying" part is done with the minimal data transformation - incoming data is being passed via "stream" and the response is being returned "as is" to the caller
- Cookies are being passed to the Falcon-Server through Falcon-Client (by sending all the required "cookie" headers)

### DIFF
--- a/packages/falcon-client/src/server.js
+++ b/packages/falcon-client/src/server.js
@@ -3,7 +3,6 @@ import serve from 'koa-static';
 import helmet from 'koa-helmet';
 import Router from 'koa-router';
 import compress from 'koa-compress';
-import url from 'url';
 import Logger from '@deity/falcon-logger';
 import error500 from './middlewares/error500Middleware';
 import serverTiming from './middlewares/serverTimingMiddleware';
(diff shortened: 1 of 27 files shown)


Borales:
@tutok we don't, but for what? having this thing "on" will make sure our app does not depend on the external env/arch + for local testing it's important too.

packages/falcon-client/src/server.js
```diff
@@ -22,6 +23,11 @@ export function Server({ App, clientApolloSchema, bootstrap, webpackAssets, load
 
   const publicDir = process.env.PUBLIC_DIR;
   const router = new Router();
+
+  if (config.graphqlProxy) {
+    graphqlProxy(config, router, port);
```
tutok: I think it should be possible to do it in that way:
```suggestion
    graphqlProxy(config, router, port);
    router.all(...graphqlProxy(config, port))
```

PR #348, opened 2019-03-13
Fix for ApiDataSource.gqlServerConfig Incorrect typing

### LINKED ISSUE #315: Incorrect typing for ApiDataSource.gqlServerConfig
**Describe the bug**
The type annotation for the `gqlServerConfig` field of the `ApiDataSource` appears to be incorrect. 

https://github.com/deity-io/falcon/blob/2e64dedc2c04df24721aa580093c7aa999c131c4/packages/falcon-server-env/src/models/ApiDataSource.ts#L47

Currently it is typed as `GraphQLSchema`, but at runtime the value is of type:

```
{
  cache: Cache;
  tracing: boolean;
  playground: object;
  schema: GraphQLSchema;
  dataSources: Function;
  formatError: Function;
  context: Function';
}
```

**To Reproduce**
Just put a breakpoint on [this line](https://github.com/deity-io/falcon/blob/dev/packages/falcon-magento2-api/src/index.js#L48) of the magento2 api provider. In fact, just reading the code shows that the type is incorrect, since it references the `schema` property.

**Expected behavior**
From what I can figure out of the Falcon internals, it looks like the correct type should be whatever the type is returned by the [`ExtensionContainer.createGraphQLConfig()`](https://github.com/deity-io/falcon/blob/dev/packages/falcon-server/src/containers/ExtensionContainer.js#L136) method. Unfortunately this is not written in TypeScript so currently is typed as `any` elsewhere.

tutok:
fix #315

### DIFF
--- a/packages/falcon-server-env/src/models/ApiDataSource.ts
+++ b/packages/falcon-server-env/src/models/ApiDataSource.ts
@@ -1,6 +1,7 @@
 import * as Logger from '@deity/falcon-logger';
 import { Body, Request, RESTDataSource } from 'apollo-datasource-rest/dist/RESTDataSource';
-import { URL, URLSearchParams, URLSearchParamsInit } from 'apollo-server-env';
+import { URLSearchParams, URLSearchParamsInit } from 'apollo-server-env';
+import { KeyValueCache } from 'apollo-server-caching';
 import { GraphQLResolveInfo, GraphQLSchema } from 'graphql';
 import { EventEmitter2 } from 'eventemitter2';
 import { stringify } from 'qs';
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1201,80 +1201,80 @@
     chalk "^2.0.1"
     slash "^2.0.0"
 
-"@jest/core@^24.4.0":
-  version "24.4.0"
-  resolved "https://registry.npmjs.org/@jest/core/-/core-24.4.0.tgz#ec53510f19dde5aaf4e9e28a3c3426fc0f2a41d6"
-  integrity sha512-S48krBwigVm3DwLSEtMiiWnWz+G3uGii192LIZYbWULYSOCwQeG7hWb6a3yWBLYuZnATh3W6QMxs7whS0/hQMQ==
+"@jest/core@^24.5.0":
```
Borales: @tutok same here with dependencies

packages/falcon-server-env/src/models/ApiDataSource.ts
```diff
@@ -14,28 +15,40 @@ import {
...
   FetchUrlResult,
-  PaginationData
+  PaginationData,
+  DataSources
 } from '../types';
 
 export type PaginationValue = number | string | null;
 
-export interface GqlServerConfig {
+export interface GqlServerConfig<TContext = any> {
```
Borales: @tutok seems like we could import Apollo's types for this type (`import { Config } from 'apollo-server-core'`) to avoid double-definition and future desync

PR #349, opened 2019-03-13
engines configuration fo CFA and example

### LINKED ISSUE #151: Add "engines" config to create-falcon-app to set the required Node ver
Add "engines" config to `create-falcon-app/package.json` file

https://docs.npmjs.com/files/package.json#engines

tutok:
resolve #151


packages/create-falcon-app/package.json
```diff
@@ -11,6 +11,10 @@
   "scripts": {
     "prepare": "node scripts/prepublish.js"
   },
+  "engines": {
+    "node": ">=8.10",
+    "yarn": ">=1.12"
```
Borales: @tutok yarn is required for local development only

Issue #357, opened 2019-03-18
Normal Module Override does not work 


tutok:
**Describe the bug**
Module Override in `falcon-client` does not work.

**To Reproduce**
1. Follow documentation, and configure a new module to override
2. Run `yarn start`
3. `falcon-client` web server response is different than page rendered via client code

**Expected behavior**
`document` returned from `falcon-client` web server should be the same as document rendered in the web browser

**Additional context**
I look like it is connected with the configuration of `webpack-node-externals` plugin

PR #358, opened 2019-03-18
Fix module override

tutok:
fix: #357

I propose to totally remove `nodeExternals`, reasons:
- it works differently in and outside of monorepo
- I have hard times when trying to configure it to compile only `@deity` packages
- `falcon-client` compilation output is not a library, so every dependency could be compiled in output bundle

I see only one disadvantage of this solution:
- the compilation time could increase. I wrote "it could" as I do not notice.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1201,32 +1201,32 @@
     chalk "^2.0.1"
     slash "^2.0.0"
 
-"@jest/core@^24.4.0":
-  version "24.4.0"
-  resolved "https://registry.npmjs.org/@jest/core/-/core-24.4.0.tgz#ec53510f19dde5aaf4e9e28a3c3426fc0f2a41d6"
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -1201,80 +1201,80 @@
     chalk "^2.0.1"
     slash "^2.0.0"
 
-"@jest/core@^24.4.0":
-  version "24.4.0"
-  resolved "https://registry.npmjs.org/@jest/core/-/core-24.4.0.tgz#ec53510f19dde5aaf4e9e28a3c3426fc0f2a41d6"
-  integrity sha512-S48krBwigVm3DwLSEtMiiWnWz+G3uGii192LIZYbWULYSOCwQeG7hWb6a3yWBLYuZnATh3W6QMxs7whS0/hQMQ==
+"@jest/core@^24.5.0":
```
Borales: @tutok I'd suggest avoiding to update any non-related packages within the same PR (that are our direct dependencies) to avoid any future unexpected surprises:

- loadable
- apollo packages
- jest

tutok: 👍

PR #366, opened 2019-03-20
Payments tweaks

mateuszwozniak:
This pull request contains couple of tweaks to the payments that we've discussed few days ago:
- removing `PaymentSelector.js` and introducing simple `TwoStepWizard` that handles any kind of wizard (can be reused for shipping and similar thing that require "pick and configure" approach)
- removing `Payment.tsx`
- rendering payment components explicitly in `PaymentMethodSection.js`
- removing component injection pattern from payment plugins - rendering payment plugin UI explicitly
- renaming:
  - `Plugin.tsx` to `PaymentPlugin.tsx`
  - `PluginModel` to `PaymentPluginModel`
  - `PluginComponentProps` to `PaymentPluginComponentProps`

All the changes here are subject of further discussion, so feel free to comment and disagree :)


examples/shop-with-blog/client/src/pages/shop/Checkout/PaymentMethodSection.js
```diff
@@ -1,12 +1,26 @@
...
 import { I18n, T } from '@deity/falcon-i18n';
-import loadable from 'src/components/loadable';
+import { ConfigQuery, TwoStepWizard } from '@deity/falcon-ecommerce-uikit';
+import loadable from '@loadable/component';
 import SectionHeader from './CheckoutSectionHeader';
 import ErrorList from '../components/ErrorList';
+import CreditCard from '../components/CreditCard';
 
-const PaymentSelector = loadable(() => import(/* webpackChunkName: "shop/payments" */ './PaymentSelector'));
+const AdyenCCPlugin = loadable(() => import(/* webpackChunkName: "checkout/payments" */ '@deity/falcon-adyen-plugin'));
```
Borales: @mateuszwozniak previously, the whole `PaymentSelector` component was loaded via loadable package (including payments), but I don't think there's gonna be some "profit" of loading each payment via the same chunk. Maybe we should load this current component (`PaymentMethodSection.js`) via loadable then?

mateuszwozniak: Yes, I don't mind, I was thinking about that already. In fact, I think that it's really enough to load the whole checkout as one chunk. At the moment I feel that we're trying to split it in too many places and it's not going to provide any visible advantage. What do you think?

Borales: @mateuszwozniak that's fine too :) we already load Checkout page via loadable, so loading separate plugins via loadable won't bring any performance improvements 👍

mateuszwozniak: done

PR #374, opened 2019-03-27
Cache for ApiDataSource

### LINKED ISSUE #373: Enable HTTPCache for ApiDataSource

Borales:
Fixes #373 

This PR introduces `@cache` GQL directive to cache your resolver results:

- `menu`
- `category`
- `product` (main data)
- `product.breadcrumbs` (via a potentially dedicated REST resolver defined dynamically)
- `url`

> By default Falcon-Server uses InMemoryCache as storage, meaning every Falcon-Server start will have an empty cache


packages/falcon-server-env/src/cache/Cache.ts
```diff
@@ -5,7 +5,7 @@ export type SetCacheOptions = {
   ttl?: number;
 };
 
-export type CacheResult = string | undefined;
+export type CacheResult = string | any | undefined;
```
tutok: `any` ? so for what we have `string` and `undefined`?

```suggestion
export type CacheResult = any;
```

packages/falcon-server/src/schemaDirectives/CacheDirective.js
```diff
@@ -0,0 +1,91 @@
+const { SchemaDirectiveVisitor } = require('graphql-tools');
+const { defaultFieldResolver } = require('graphql');
+const crypto = require('crypto');
+
+// Default cache TTL (10 minutes)
+const DEFAULT_TTL = 10;
```
tutok: why `10` minutes? should we have even had default value configured? So if I understand correctly it is not possible to skip caching until I specify it explicitly in the config [here](https://github.com/deity-io/falcon/pull/374/files#diff-f367dacb9ba5614b667ceb3e06ebef72R96)  `"cache": { "default": { "ttl": 0 }` ?

For me looks too obtrusive. Personally, I would expect default value to be set to `0` - no caching. If I decide that 10 min works for me then I would like to set it in config `"cache": { "default": { "ttl": 10 }`

packages/falcon-server-env/src/cache/Cache.ts
mateuszwozniak: true that :)

packages/falcon-server/src/schemaDirectives/CacheDirective.js
```diff
@@ -0,0 +1,91 @@
...
+  getResolverWithCache(resolve, field, defaultValue) {
+    const thisDirective = this;
+    return async function fieldResolver(parent, params, context, info) {
+      const {
+        config: { cache: cacheConfig = {} }
+      } = context;
+      const { ttl } = thisDirective.getCacheConfigForField(info, cacheConfig, defaultValue);
+      const { name: fieldName } = field;
+      // Generating short and unique cache-key
+      const cacheKey = crypto
```
tutok: should we waste resources for calculating `cacheKey` (and latter use `context.cache.get`) in case if `ttl` is `0`

packages/falcon-server/src/schemaDirectives/CacheDirective.js
```diff
@@ -0,0 +1,91 @@
...
+   * @param {object} cacheConfig Cache object provided via `context.config`
+   * @param {object} defaultDirectiveValue Default options defined in cache directive for the specific type
+   * @return {object} Final cache options object
+   */
+  getCacheConfigForField(info, cacheConfig, defaultDirectiveValue) {
+    const { path: gqlPath, operation } = info;
+
+    // Generates a path-like string for the provided request
+    // for `query { foo { bar } }` - it will generate "foo.bar" string
+    const getFullPath = node => {
```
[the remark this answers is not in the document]
Borales: moved

PR #392, opened 2019-04-05
Set cursor pointer on hover product custom option

yogeshsuhagiya:

- Bug fix

- Default cursor appears on hover product custom options like color, size, etc.

- Set cursor pointer on hover product custom options.

- No

**Other information:**


cane-lagerwaard:
Hi @yogeshsuhagiya thanks for your PR :) We'll take a look at it asap

packages/falcon-ui/src/components/Radio.tsx
```diff
@@ -64,6 +64,9 @@ export const Radio = themed({
               fill: theme.colors.primary
             }
           },
+          ':hover': {
+            cursor: 'pointer'
```
tutok: we can not set `pointer` for all `radio` `inputs` as `falcon-ui` provides low level UI components and default value is `default`.

value of this property should(or could) be controller via props, and actually on product option it is done [here](https://github.com/deity-io/falcon/blob/eba32bff847ad19fedae15bc7b9b04584a2abf46/packages/falcon-ecommerce-uikit/src/Product/ConfigurableOptions.tsx#L17), which is good.

so I think better (but still not correct) solution for this issue could be:
```suggestion
            cursor: 'inherit'
```

yogeshsuhagiya: Hi @tutok, yes it works. I've checked your solution in my local project and changed accordigly. Thanks for correcting me. :)

packages/falcon-ui/src/components/Radio.tsx
```diff
@@ -64,6 +64,9 @@ export const Radio = themed({
               fill: theme.colors.primary
             }
           },
+          ':hover': {
```
tutok: Please forgive me for my minuteness, but I belive pseudo selector `:hover` is not necessary here
```suggestion
```

second thing is, if we set `cursor: 'inherit'` for `input` then we need to set `cursor: 'default'` on its parent to make sure that we do not brake default look and fill in other place.

So on the end I would rather expect something like this:

```
  radio: {
    // ...
    css: ({ theme }) => ({
      // ...
      cursor: 'default',
      input: { 
        cursor: 'inherit',
        // ...
      }
      // ...
    })
  }
```

thanks in advance for your help!

yogeshsuhagiya: Hi @tutok 
I've updated CSS property as per your expectation, please review it and don't hesitate to ask for changes.

tutok:
@yogeshsuhagiya great job 👍, thanks!

Issue #399, opened 2019-04-10
After successful checkout user is not able to pick his address when trying to do another checkout


mateuszwozniak:
**Describe the bug**
If you try to go through the checkout twice (without website reload) then you won't be able to select shipping/billing address from your address book during second checkout process. 

**To Reproduce**
1. Log in
2. Make sure that you have at least one address in your address book 
3. Add something to cart
4. Go to checkout (you can see that shipping and billing sections display address picker with your addresses)
5. Finalize purchase
6. Add new thing to the cart
7. Go to checkout page - there's no address picker in shipping and billing sections - you have to provide new address

**Expected behavior**
You should be able to pick shipping and billing address from your addresses

tutok:
I am not sure it is connected to this issue but looks like it could be:

**To Reproduce**
Steps to reproduce the behavior:
1. Log in
2. Add something to cart
3. Go to checkout, you should be able to proceed it as Logged In user (you can see that shipping and billing sections display address picker with your addresses)
5. Refresh the page (press `f5`)
6. Checkout does not recognize yours Logged In state

mateuszwozniak:
@tutok, you're right. It looks like it's related to the same issue.

PR #404, opened 2019-04-10
Fix checkout issue

mateuszwozniak:

There are 2 changes:
1. Removed workaround (introduced in commit 9b16c43) that was making sure that CheckoutLogic does not throw an error when ApolloClient calls success callback twice. New version of ApolloClient does not have that problem so the workaround is not needed anymore
2. Fixed name of id attribute in magento2 api - it's called now `customer_address_id` as magento accepts that one (instead of `id` when sending address from address book)

See #399.

The problem is removed - user can now correctly use address from his address book.

### DIFF
--- a/packages/falcon-magento2-api/src/index.js
+++ b/packages/falcon-magento2-api/src/index.js
@@ -1648,12 +1648,18 @@ module.exports = class Magento2Api extends Magento2ApiBase {
   }
 
   /**
-   * Removes unnecesary fields from address entry so Magento doesn't crash
+   * Removes unnecessary fields from address entry and adds proper id so Magento doesn't crash
    * @param {AddressInput} address - address to process
    * @return {AddressInput} processed address
    */
   prepareAddressForOrder(address) {
     const data = { ...address };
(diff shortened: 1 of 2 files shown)


packages/falcon-magento2-api/src/index.js
```diff
@@ -1680,6 +1686,8 @@ module.exports = class Magento2Api extends Magento2ApiBase {
 
     const falconPrefix = FALCON_CART_ACTIONS.indexOf(path) === -1 ? '' : '/falcon';
 
+    console.log(method.toUpperCase(), `${falconPrefix}${cartPath}${path}`, JSON.stringify(data, null, 2));
```
Borales:
```suggestion
```

tutok:
```suggestion
```
should not we remove that line?

mateuszwozniak: omg, sorry :)

mateuszwozniak: removed :)

PR #420, opened 2019-04-16
falcon scripts

tutok:

introduced `@deity/falcon-pkg` with package building scripts

N/A

N/A

no

**Other information:**
N/A


package.json
```diff
@@ -55,6 +55,7 @@
     "packages/falcon-dev-tools/babel-preset-falcon-client",
     "packages/falcon-dev-tools/eslint-config-falcon",
     "packages/falcon-dev-tools/falcon-i18n-webpack-plugin",
+    "packages/falcon-dev-tools/falcon-pkg",
```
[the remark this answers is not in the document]
tutok: I am aware that the name `falcon-pkg` is not the best, personally, I would rather prefer `falcon`, but it is too wide for having in mind current situation (we have already `falcon-client` and `falcon-server` which is not used as I can see). So I chose `falcon-package` (as scripts can be applied only to packages) and later make it shorter to `falcon-pkg`.

About suggested names.  If I need to choose from them, I think that `@deity/env` is better. But I will wait till @mateuszwozniak answer as he also was mentioned above, and I would not like to change the name back and forth.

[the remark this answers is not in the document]
mateuszwozniak: I'd call it `@deity/build-tools` or `@deity/development-tools` or something with `tools` as `@deity/env` doesn't say too much. `@deity/env-tools` is better than `@deity/env` but the `env` itself isn't the best choice imo. Maybe just `falcon-scripts` in a similar manner to `react-scripts`? In that case entry in package.json would be:
```
"watch": "falcon-scripts watch" 
``` 
and that looks pretty good imo.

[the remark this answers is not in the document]
Borales: `@deity/falcon-scripts` looks good to me 👍

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 Versions marked with a number and date (e.g. Falcon Client v0.1.0 (2018-10-05)) are already released and available via npm. Versions without a date are not released yet.
 
+## Falcon vNEXT
+
+### Falcon Pkg v0.0.1
```
Borales: @tutok please change this line too
```suggestion
### Falcon Pkg v0.0.1
```

tutok: THANKS! 👍

packages/falcon-ecommerce-uikit/package.json
```diff
@@ -106,25 +69,17 @@
...
-    "lodash.isequal": "^4.5.0",
-    "qs": "6.6.0",
-    "react-adopt": "^0.6.0",
-    "react-credit-card-input": "1.1.1",
-    "react-lazyload": "2.5.0",
-    "react-powerplug": "1.0.0",
-    "react-responsive": "6.1.1"
+  "peerDependencies": {
+    "@deity/falcon-i18n": "0.0.3",
+    "@deity/falcon-ui": "0.0.1",
```
Borales: @tutok previously, there were:

```
    "@deity/falcon-i18n": "^0.0.3",
    "@deity/falcon-ui": "^0.0.1",
```

We shouldn't be changing `^` for these packages since Lerna doesn't seem to be updating them with the new versions

tutok: done

packages/falcon-theme-editor/package.json
```diff
@@ -4,56 +4,40 @@
...
-    "rollup-plugin-node-resolve": "3.4.0",
-    "typescript": "3.2.1"
+    "react-dom": "^16.6.3"
   },
-  "dependencies": {
-    "@babel/runtime": "7.3.1",
-    "deep-object-diff": "^1.1.0",
-    "webfontloader": "^1.6.28"
+  "peerDependencies": {
+    "@deity/falcon-ui": "^0.0.2",
```
Borales:
```suggestion
    "@deity/falcon-ui": "^0.2.1",
```

PR #440, opened 2019-05-12
Cache invalidation

### LINKED ISSUE #414: Invalidate cache by tags
Implement "cacheMutators" to trigger cache invalidation for certain tags after completing the Mutation operation (like `placeOrder`)

### LINKED ISSUE #423: Force fetch on a client-side for critical data
The current default behavior of ApolloClient is that once it fetches the data - all further client-side calls simply reuse the data from the state. In some cases - that can be acceptable (list static posts), but for critical data, we have to make sure we use remote data all the time.

Critical data:
- Product (including stock status, visibility, price etc)
- Product listings

To avoid performance downgrade - we should use `cache-and-network` [fetch policy](https://www.apollographql.com/docs/react/api/react-apollo#graphql-config-options-fetchPolicy), so the component would render the data from the state and then make a request to the Falcon-Server to get the latest state.

Borales:

**Other information:**
Fixes #423 
Resolves #414


packages/falcon-server/src/graphqlUtils/index.js
```diff
@@ -0,0 +1,155 @@
...
+  });
+};
+
+/**
+ * Generate tag names using `entityName` and `entityId`
+ * @param {string} entityName Entity Type name (like "Product")
+ * @param {string|string[]} entityId Entity ID or list of IDs (like: "1" or ["1", "2"])
+ * @return {string[]} List of tag names (example: ["Product:1", "Product:2"])
+ */
+const generateTagNames = (entityName, entityId) => {
```
tutok: It looks like these functions (`generateTagNames`, `getTagsForField` and `extractTagsForIdPath`) are more cache specific than graphQL, others could be potentially moved into our  future `utils` package. Am I right? I am asking because I would like to know where the division line is placed (if it is defined already).

packages/falcon-server/src/graphqlUtils/index.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+  const ids = Array.isArray(entityId) ? entityId.filter(element => element) : [entityId];
+  return ids.map(id => `${entityName}${TAG_SEPARATOR}${id}`);
+};
+
+/**
+ * Get a list of tags from the provided `sourceValue` using specified `fieldType` and `fieldPathSections` (for nested values)
+ * @param {object} sourceValue Source value to get tags from
+ * @param {GraphQLType} fieldType GraphQL Field Type object
+ * @param {string[]} [fieldPathSections=[]] An optional field path sections (example: ["products", "items"] which are created from a relative "products.items" field path)
```
tutok:
```suggestion
 * @param {string[]} [fieldPathSections=[]] An optional field path sections (example: `["products", "items"]` which are created from a relative `products.items` field path)
```

packages/falcon-server/src/middlewares/cacheMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+ * @return {Middleware} Koa middleware callback
+ */
+const cacheMiddleware = cache => async ctx => {
+  /** @type {Array<CacheTagEntry>} List of posted cache tag entries to invalidate */
+  const requestTags = ctx.request.body;
+  if (ctx.request.get('content-type') !== 'application/json') {
+    throw new Error('Invalid Content-Type, must be "application/json"');
+  }
+  if (!Array.isArray(requestTags)) {
+    throw new Error('Invalid POST data');
```
tutok: I think we should return `400`http error code here, am I right?

tutok: I think we should return `400`http error code here, am I right?

packages/falcon-server/src/middlewares/cacheMiddleware.js
```diff
@@ -0,0 +1,39 @@
...
+ * @property {string} id Entity ID
+ */
+
+/**
+ * Cache middleware for handling web-hooks to flush the cache by tags
+ * @example curl -X POST http://localhost:4000/cache -H 'Content-Type: application/json' -d '[{"id": 1, "type": "Product"}]'
+ * @param {Cache} cache Cache component
+ * @return {Middleware} Koa middleware callback
+ */
+const cacheMiddleware = cache => async ctx => {
```
mateuszwozniak: A better name would be `cacheInvalidationMiddleware` or something like that as that's its purpose

packages/falcon-server/src/graphqlUtils/index.js
mateuszwozniak: 👍

packages/falcon-server/src/middlewares/cacheMiddleware.js
Borales: @tutok updated

packages/falcon-server/src/graphqlUtils/index.js
Borales: @tutok 

- `generateTagNames` - it can be moved as it's totally generic
- `getTagsForField` -  we expect `GraphQLType` argument type to be passed to this function, so it's still related to the GQL utils
- `extractTagsForIdPath` - same thing with `GraphQLResolveInfo` argument

but since `generateTagNames` is being used by `getTagsForField` internally - I'd consider keeping it where it is now (not really sure it makes such a difference).

PR #453, opened 2019-05-22
Fixing default storeCode

Borales:


Borales:
> Can you please add also `defaultStoreCode` to `default.json` in the example?

@mateuszwozniak done

Issue #419, opened 2019-04-16
[Discussion] ServiceWorker improvements


Borales:
### What should be changed/improved?

The current behavior of ServiceWorker within Falcon-Client context could be improved. 
With the latest version of Workbox package (latest - `4.3.0`, we use - `3.5.0`) - workbox does not allow us to call `workbox.skipWaiting()` method anymore (like we did in https://latest.snapshot.deity.io/sw.js), which forces browsers to update it immediately (applying the latest serviceWorker instance).

[ServiceWorker lifecycle](https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle) discourages such behavior:

> - Allow a new service worker to get itself ready without disrupting the current one.
> - Ensure an in-scope page is controlled by the same service worker (or no service worker) throughout.
> - Ensure there's only one version of your site running at once.
> 
> That last one is pretty important. Without service workers, users can load one tab to your site, then later open another. This can result in two versions of your site running at the same time. Sometimes this is ok, but if you're dealing with storage you can easily end up with two tabs having very different opinions on how their shared storage should be managed. This can result in errors, or worse, data loss.

The current behavior could potentially harm user-experience for the visitors, and therefore decrease the shop revenue.

### Your suggested solution

- Falcon-Client's implementation of ServiceWorker generation should not be forcing browsers to apply ServiceWorker immediately (through `skipWaiting`), but rather follow the proper ServiceWorker lifecycle
- ServiceWorker must be able to precache our bundle files correctly, and to make it possible - we need to get rid of "hash" suffix in bundle filenames (`/static/js/client.0f3ef189.js` must be `/static/js/client.js`, so workbox itself would calculate the revision automatically). This way - ServiceWorker will be able to calculate the correct "diff" for the precached files by matching filenames from `sw.js` and "Cache storage" items (so deploying new version - won't break the UI by providing new bundle files)
- Allow developers to define their own custom SW template (for example `client/src/sw.js`) so Falcon-Client could check it or apply the built-in one.

### Things to check/investigate

- Server-Side Rendering while keeping the "previous" ServiceWorker and precached bundle files

### TODO

- [x] Update Workbox package to the [latest version](https://github.com/GoogleChrome/workbox/releases)
- [x] Remove hashes from bundle filenames
- [x] Allow developers to define their own SW template

---

This is a "ticket" to discuss this improvement openly to get opinions from different perspectives.

CC: @mateuszwozniak @tutok @cane-lagerwaard 

---

[...]

PR #454, opened 2019-05-23
Custom sw

tutok:

it will be possible to inject custom code into `sw.js` generated via `falcon-clinet`. To do this you need to create file `your-example-shop-with-blog-app\client\sw.js` 

no

**Other information:**
no


packages/falcon-client/build-utils/webpack/index.js
```diff
@@ -4,6 +4,7 @@ const Logger = require('@deity/falcon-logger');
 const WebpackDevServer = require('webpack-dev-server-speedy');
 const clearConsole = require('react-dev-utils/clearConsole');
 const { measureFileSizesBeforeBuild, printFileSizesAfterBuild } = require('react-dev-utils/FileSizeReporter');
+const serviceWorker = require('./../serviceWorker');
```
mateuszwozniak:
```suggestion
const serviceWorker = require('../serviceWorker');
```

packages/falcon-client/build-utils/webpack/index.js
```diff
@@ -4,6 +4,7 @@ const Logger = require('@deity/falcon-logger');
...
 
 const paths = require('./../paths');
```
mateuszwozniak:
```suggestion
const paths = require('../paths');
```

packages/falcon-client/build-utils/serviceWorker/index.js
```diff
@@ -0,0 +1,64 @@
...
+        resolve(),
+        re({
+          patterns: [
+            {
+              test: 'const ENTRIES = [];',
+              replace: `const ENTRIES = ${JSON.stringify(manifestEntries, null, 2)};`
+            }
+          ]
+        })
+        // isProductionBuild ? terser() : undefined
```
mateuszwozniak:
```suggestion
```

packages/falcon-client/sw.js
```diff
@@ -0,0 +1,2 @@
+// re-exporting falcon-client service worker defaults
```
Borales: @tutok what does "re-exporting" mean in this context?

packages/falcon-client/src/serviceWorker/pre-cache.js
```diff
@@ -0,0 +1,20 @@
+/* global workbox, self, importScripts */
+
+importScripts('https://storage.googleapis.com/workbox-cdn/releases/3.5.0/workbox-sw.js');
```
Borales: @tutok just a question - how hard would it be to upgrade workbox to the latest ver?

packages/falcon-client/src/serviceWorker/sw.js
```diff
@@ -1,18 +1 @@
...
-workbox.precaching.suppressWarnings();
-workbox.precaching.precacheAndRoute([], {});
-
-workbox.routing.registerRoute(
-  ({ event }) => event.request.mode === 'navigate',
-  ({ url }) => fetch(url.href).catch(() => caches.match('/app-shell'))
-);
-
-workbox.routing.registerRoute('/', workbox.strategies.networkFirst(), 'GET');
+import './pre-cache';
```
Borales: @tutok can we keep the content of this file as it was before without re-importing it? if people would be looking for the original template - they would have to jump across re-imported files, feels like an unnecessary thing to do.

packages/falcon-client/build-utils/serviceWorker/index.js
```diff
@@ -0,0 +1,52 @@
...
+      sourcemap: !isProductionBuild,
+      compact: isProductionBuild
+    };
+
+    const bundle = await rollup.rollup(inputOptions);
+    await bundle.write(outputOptions);
+  } catch (error) {
+    Logger.error(`${chalk.red('\nFailed to compile.\n')}`);
+    Logger.error(error);
+    Logger.log();
```
Borales:
```suggestion
```

packages/falcon-client/src/serviceWorker/sw.js
tutok: I can move it back, but is this means that we should put entire code base into a single file?

packages/falcon-client/sw.js
tutok: I belive that we should bot expose our internals, so instead 
```
import '@deity/falcon-client/src/serviceWorker/sw';
```
I done 
```
import '@deity/falcon-client/sw';
```

which is also rather easy to convert into separate package `@deity/falcon-client-sw` in future

packages/falcon-client/build-utils/serviceWorker/index.js
tutok: removing this line will cause that output will do not look so nice anymore, are you sure you want this?

packages/falcon-client/src/serviceWorker/sw.js
Borales: @tutok it will be imported into the project's `sw.js` anyways, right? 

We still store the whole code for `sw` template in a single file (`pre-cache.js`), so this `sw.js` simply re-imports/re-exports it.

packages/falcon-client/src/serviceWorker/pre-cache.js
tutok: I think that it will be as hard as mainly rethinking out entire pre-caching approach because they change it totally since [version `4.0`](https://github.com/GoogleChrome/workbox/releases/tag/v4.0.0-alpha.0)

I believe #419 was originally created to do this, at least to discuss new approach.

packages/falcon-client/build-utils/serviceWorker/index.js
Borales: @tutok hm... ok  😄

packages/falcon-client/src/serviceWorker/sw.js
Borales: What I meant is that we have `falcon-client/sw.js` file that acts as an "entrypoint" for the import call inside of a project's `sw.js`. To make sure we won't expose all the things - we have this `falcon-client/src/serviceWorker/sw.js` file which stores all the internal business logic for SW. So my question was about having another separate file to deliver the code.

tutok: like you already may realize my thinking was different,
* `falcon-client/sw.js` exists because of package export
* `falcon-client/src/serviceWorker/sw.js` exists because of the entry point

but adjusted as you requested

packages/falcon-client/bin/falcon-client.js
```diff
@@ -9,24 +9,25 @@ process.on('uncaughtException', ex => {
 process.noDeprecation = true; // turns off that loadQuery clutter.
 
 const Logger = require('@deity/falcon-logger');
-const { start, build, size, test } = require('./../build-utils');
+const { app, serviceWorker, test } = require('./../build-utils');
```
mateuszwozniak:
```suggestion
const { app, serviceWorker, test } = require('../build-utils');
```

tutok: done

PR #468, opened 2019-05-31
Configured TypeScript in Magento2 API 

tutok:

added TypeScript into `@deity/falcon-magento2-api` package

No

**Other**
this PR does not introduce TS everywhere, but it is a good starting point at least.


packages/falcon-magento2-api/src/index.ts
```diff
@@ -1,17 +1,19 @@
-const url = require('url');
```
nimobeeren: Getting a type error on the following line (can't place a comment there unfortunately), it's expecting a promise to be returned:

https://github.com/deity-io/falcon/blob/c814cf32f083b593a73b0cb770acaf3435065710/packages/falcon-magento2-api/src/index.ts#L651

Maybe just making it `async` would be a nice solution?

packages/falcon-magento2-api/src/authorization.ts
```diff
@@ -1,10 +1,11 @@
-const deepMerge = require('deepmerge');
-const OAuth = require('oauth');
+import deepMerge from 'deepmerge';
+import OAuth from 'oauth';
```
nimobeeren:
```suggestion
import { OAuth } from 'oauth';
```

and on this line:

https://github.com/deity-io/falcon/blob/c814cf32f083b593a73b0cb770acaf3435065710/packages/falcon-magento2-api/src/authorization.ts#L75

replace with `new OAuth(...)`

packages/falcon-magento2-api/src/index.ts
tutok: it is weird, I have run `yarn install`, `yarn lint`  and no errors, also in IDE. I am using TS in version `3.6.4`, also I do not realize runtime errors.

packages/falcon-magento2-api/src/authorization.ts
tutok: done

Borales:
@tutok After installing repo's dependencies and running example app - getting this error. Could you please take a look at it?

Screenshot 2019-11-04 at 11 00 44

packages/falcon-server-env/src/types.ts
```diff
@@ -44,7 +44,7 @@ export interface IConfigurableConstructorParams<T = object> {
 
 export type ContextType = {
   isAuthRequired?: boolean;
-  didReceiveResult?: (result: any, res: ContextFetchResponse) => Promise<any>;
```
Borales: @tutok I'd keep the previous typing, as `didReceiveResult` needs to return a Promisified result for:

```
protected async didReceiveResponse<TResult = any>(res: ContextFetchResponse, req: Request): Promise<TResult> {
   return res.context.didReceiveResult(result, res);
}
```

even though `any` covers this part, it needs to be a Promisified `any` (`Promise<any>`)

packages/falcon-magento2-api/package.json
```diff
@@ -7,13 +7,23 @@
...
+  ],
   "jest": {
     "setupFiles": [
       "../../test/nodeEnvSetupFile"
     ]
   },
   "scripts": {
+    "prepare": "npm run clean && npm run build",
+    "clean": "falcon-scripts clean",
+    "build": "falcon-scripts build --target=NODE",
```
Borales:
```suggestion
    "build": "falcon-scripts build",
```

packages/falcon-magento2-api/package.json
```diff
@@ -7,13 +7,23 @@
...
+    "watch": "falcon-scripts watch --target=NODE",
```
Borales:
```suggestion
    "watch": "falcon-scripts watch",
```

packages/falcon-magento2-api/src/index.ts
nimobeeren: I switched to 3.6.4 and back to 3.6.3 and the error was gone :upside_down_face:. You can ignore this I guess :smile:

nimobeeren:
Also getting this error @Borales mentioned on node v13.0.1

packages/falcon-magento2-api/src/index.ts
tutok: @nimobeeren please try one more time

nimobeeren: IDE and lint give no errors, looks all good to me.

One weird thing is that running `yarn lint` in root directory does not detect any type errors at all, even when I remove `async` there, while IDE does detect it.

Issue #436, opened 2019-05-08
RFC: Re-work of falcon-ecommerce-uikit library


mateuszwozniak:
The purpose of this issue is to plan rework of the `@deity/falcon-ecommerce-uikit` package. The initial idea was to have a set of UI components that help building UI for shops. Unfortunately over the time we put almost everything related to front-end of shop in there - Queries, Mutations, higher order components that provide logic without rendering UI, reusable UI pieces and even full page views.

Because of that it's very hard to maintain that library and also provide enough flexibility so developers are not forced to use particular things.

Besides architectural issues there are a lot of issues with naming (mentioned at the end of this description) that we need to solve during this word.

> NOTE: the term "business level components" in this issue relates to the components that do not render UI by themselves but provide a piece of logic that's required to achieve a particular functionality. That mostly covers Higher Order Components and Render Prop Components.

The list with everything that's inside `@deity/falcon-ecommerce-uikit` is [here](https://github.com/deity-io/falcon/blob/uikit-rework/packages/falcon-ecommerce-uikit/uikit.md). Before adding comments here please get familiar with that list as it's an entrypoint to the discussion about potential changes. Once you go through that list you'll understand why we have to the changes listed below :)

# 1. Goals to achieve:

1. Organizing the code into separate logical packages (UI, business logic, data access)
2. Splitting the code to small units (component per file approach?) so component overriding works at all levels
3. Providing documentation for everything (listing components and their purpose)
4. Moving pieces that should not be inside separate libraries to the example shop
5. Unifying naming convention
6. Produce conventions/standards for naming:

- gql constants
- Query and Mutation components
- data types
- "business level" components
- UI components

## NOTES

- `RenderProps` vs `InjectedProps` (e.g. `EnsureTTIRenderProps` and `CheckoutLogicInjectedProps`)
- `Logic` vs `Provider` (e.g. `CheckoutLogic` and `FiltersDataProvider`)
- `SomethingLayout` components (created usually with `themed()` call or as React.SFC) vs `itemSomethingLayout` properties - defined as props of `DefaultThemeProps` type - both have "Layout" in the name and that's confusing. Examples: `MiniFormLayout` (React.SFC), `AddressCardLayout` (`themed()`) and `orderItemSummaryLayout` (const of `DefaultThemeProps` type)
- inconsistencies between query names and its data (`OrderListQuery` returns `OrdersData`, while it should be `OrderListData`)

# 2. Suggested changes

## 2.1. Suggested packages

[...]

PR #475, opened 2019-06-07
Uikit rework - new packages

mateuszwozniak:
Initial changes related to rework of uikit package (see #436 ):
- created packages `falcon-shop-data` and `falcon-uikit`
- moved, splited and renamed Customer related data types, queries and mutations to proper places
- moved, splited and renamed Address related data types, queries, mutations and UI components to proper places
- adjusted logic for Extension Container so it's able to load '.default' component from extension's package (this is a subject of further changes so it's a temporary solution)

Feel free to comment all the work and suggest changes as that's a base for all the future work related to that refactoring so we should be sure that it's in the shape that everybody's fine with :)


packages/falcon-shop-data/src/Customer/IsAuthenticatedQuery.tsx
```diff
@@ -9,14 +9,15 @@ export const GET_IS_AUTHENTICATED = gql`
   }
 `;
 
-export type IsAuthenticatedQueryData = {
+export type IsAuthenticatedQueryResponse = {
```
tutok:
```suggestion
export type IsAuthenticatedResponse = {
```

maybe we could make it shorter. The same convention could be used for `mutations`.
Let say we have `Adderess`  and
* `query` `GET_ADDRESS_BY_ID` will return  `AddressResponse`
* `mutation` `ADD_ADDRESS` will return  `AddressResponse` too

packages/falcon-uikit/src/Address/utils.ts
```diff
@@ -0,0 +1,6 @@
+import { Address } from '@deity/falcon-shop-extension';
```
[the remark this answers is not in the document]
mateuszwozniak: Good point, I've created #481 for that

packages/falcon-shop-extension/src/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  company?: string;
+  region?: string;
+  regionId?: number;
+  telephone?: string;
+  fax?: string;
+  defaultBilling: boolean;
+  defaultShipping: boolean;
+};
+
+export type Customer = {
```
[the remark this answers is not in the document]
mateuszwozniak: Created #482 for that

packages/falcon-shop-data/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Customer';
```
tutok:
```suggestion
export * from './Address';
export * from './Customer';
```

packages/falcon-dev-tools/eslint-config-falcon/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@deity/eslint-config-falcon",
-  "version": "0.1.2",
+  "version": "0.1.1",
```
Borales: @mateuszwozniak will it break the upcoming release? (in case of a patch version)

packages/falcon-server-env/package.json
```diff
@@ -6,7 +6,7 @@
   "repository": "https://github.com/deity-io/falcon/tree/master/packages/falcon-server-env",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
-  "source": "src/index.ts",
+  "source": "src/index.js",
```
Borales: @mateuszwozniak are you sure it's correct?

packages/falcon-dev-tools/eslint-config-falcon/package.json
mateuszwozniak: sorry, fixed now

packages/falcon-server-env/package.json
mateuszwozniak: Damn, good point. I had to do that accidentally, fixed now.

packages/falcon-shop-extension/src/types.ts
```diff
@@ -0,0 +1,28 @@
+export type Address = {
```
Borales: @mateuszwozniak should this go to `shop-data`?

PR #485, opened 2019-06-12
Attempt to move Forms folder to new uikit package

tutok:
Scope:
- moved Form-related components to @deity/falcon-ui-kit
- renamed types in typing and schema files to match the convention
- added @deity/falcon-data package for data-related components
- added @deity/falcon-front-kit package for front-end "business level components"


packages/falcon-ui-kit/src/Account/ResetPassword.tsx
```diff
@@ -2,11 +2,14 @@ import React from 'react';
...
-import { FormField, Form, FormSubmit, FormErrorSummary, PasswordRevealInput } from '../Forms';
-import { ResetCustomerPasswordMutation, ResetCustomerPasswordVariables } from './AccountRecoveryMutations';
-import { ValidatePasswordTokenQuery } from './AccountRecoveryQueries';
+import { ResetCustomerPasswordMutation, ValidatePasswordTokenQuery } from '@deity/falcon-shop-data';
+import { ResetCustomerPasswordInput } from '@deity/falcon-shop-extension';
 import { OpenSidebarMutation } from '../Sidebar';
-import { MiniFormLayout } from '../MiniAccount';
+import { FormField, Form, FormSubmit, FormErrorSummary, PasswordRevealInput } from '../Forms';
+
+// todo (uikit): re-work that once we have MiniAccount and "MiniLayouts" moved to falcon-ui-kit
```
mateuszwozniak: I'm going to fix that in the next pull request as it requires moving all sidebar components (layouts, forms etc) and I didn't want to mix everything into one pull request

packages/falcon-shop-extension/src/schema.graphql
```diff
@@ -217,12 +217,12 @@ input EditAddressInput {
   regionId: Int
 }
 
-input CustomerPassword {
+input CustomerPasswordInput {
```
tutok: this type represents data required to set a new password, instead of password in general I think

```suggestion
input NewPasswordInput {
```

packages/falcon-ui-kit/src/Forms/RadioFormField.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  // eslint-disable-next-line
+  let validators = validate || [];
+  if (required) {
+    validators.unshift(requiredValidator);
+  }
+  const defaultInputTypeValidator = !validate && getDefaultInputTypeValidator(restProps.type);
+  if (defaultInputTypeValidator) {
+    validators.push(defaultInputTypeValidator);
+  }
```
tutok: I think we could improve the above code which is responsible for setting `validators`. The main reason is that is duplicated in each `*FormField`. So a new task for this? what do you think?

packages/falcon-shop-data/src/Account/RequestPasswordResetMutation.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Mutation } from 'react-apollo';
+import { EmailInput } from '@deity/falcon-shop-extension';
+
+export const REQUEST_CUSTOMER_PASSWORD_RESET_TOKEN_MUTATION = gql`
+  mutation RequestCustomerPasswordResetToken($input: EmailInput!) {
+    requestCustomerPasswordResetToken(input: $input)
+  }
+`;
+
+export type RequestPasswordResetVariables = {
```
tutok:
```suggestion
export type RequestPasswordResetTokenVariables = {
```

packages/falcon-shop-data/src/Account/RequestPasswordResetMutation.tsx
```diff
@@ -0,0 +1,19 @@
...
+  input: EmailInput;
+};
+
+export class RequestPasswordResetMutation extends Mutation<boolean, RequestPasswordResetVariables> {
```
tutok:
```suggestion
export class RequestPasswordResetTokenMutation extends Mutation<boolean, RequestPasswordResetTokenVariables> {
```

packages/falcon-shop-data/src/Account/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './RequestPasswordResetMutation';
+export * from './ResetCustomerPasswordMutation';
+export * from './ValidatePasswordTokenQuery';
```
tutok: maybe we could extract types for `ValidatePasswordTokenQuery`
```
export type ValidatePasswordTokenResponse = {
  isValid: boolean;
}
```

packages/falcon-shop-data/src/Account/RequestPasswordResetMutation.tsx
```diff
@@ -0,0 +1,19 @@
+import gql from 'graphql-tag';
+import { Mutation } from 'react-apollo';
+import { EmailInput } from '@deity/falcon-shop-extension';
+
+export const REQUEST_CUSTOMER_PASSWORD_RESET_TOKEN_MUTATION = gql`
+  mutation RequestCustomerPasswordResetToken($input: EmailInput!) {
```
tutok: I am not convinced that we should use `EmailInput` here, I think we should create `RequestCustomerPasswordResetTokenInput` type instead:
```
export type RequestCustomerPasswordResetTokenInput = {
   email: string;
}
```

```suggestion
  mutation RequestCustomerPasswordResetToken($input: RequestCustomerPasswordResetTokenInput!) {
```

packages/falcon-shop-extension/src/schema.graphql
mateuszwozniak: Following our convention it should rather be `ChangeCustomerPasswordInput` (mutation name + Input).

packages/falcon-shop-data/src/Account/index.ts
mateuszwozniak: > * we have `<FormSubmit />` which is expecting `submitting` prop, I think it should be possible get rid of that "additional" prop via connecting to current `formik` Form via https://jaredpalmer.com/formik/docs/api/connect

We could do so but that `submitting` is not always value of `isSubmitting` prop from formik. We also pass value of `loading` that we get from query/mutation as that parameter (which can have different state from form submitting when query refetching is involved) so sticking to formik's value would remove that possibility.

packages/falcon-shop-data/src/Account/RequestPasswordResetMutation.tsx
mateuszwozniak: done

packages/falcon-shop-data/src/Account/RequestPasswordResetMutation.tsx
mateuszwozniak: done

packages/falcon-shop-data/src/Account/index.ts
mateuszwozniak: > ValidatePasswordTokenQuery

Good point, fixed.

packages/falcon-shop-data/src/Account/RequestPasswordResetMutation.tsx
mateuszwozniak: yep, good point - done.

packages/falcon-ui-kit/src/Sidebar/CloseSidebarMutation.tsx
```diff
@@ -0,0 +1,14 @@
+import gql from 'graphql-tag';
+import { Mutation } from 'react-apollo';
+
+const CLOSE_SIDEBAR_MUTATION = gql`
+  mutation CloseSidebarMutation {
+    closeSidebar @client
+  }
+`;
+
+export class CloseSidebarMutation extends Mutation {
```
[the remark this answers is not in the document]
mateuszwozniak: Created #496

PR #491, opened 2019-06-17
Node source-maps support

Borales:


mateuszwozniak:
What about Falcon Client?

Borales:
@mateuszwozniak I didn't want to mix all the source-maps at once, it was a part of another task, but ok, added

tutok:
@mateuszwozniak do you mean adding 
```
require('source-map-support').install();
```
also here https://github.com/deity-io/falcon/blob/8e1030088f4477d39b7e9ab325d367499c69cc9a/packages/falcon-client/src/index.js#L1
❔

mateuszwozniak:
@tutok yes

Borales:
@mateuszwozniak added

Issue #262, opened 2019-01-15
Provide developer-friendly error page


Borales:
Whenever Falcon app has `debug: true` or `NODE_ENV: development` flags - we need to provide an extended "error page" with the whole stack trace and proper error message.

Currently, we show general "Apologies" page.

PR #494, opened 2019-06-17
Moved part of business components into falcon-front-kit

tutok:

**Other information:**


packages/falcon-data/src/BackendConfig/BackendConfigQuery.tsx
```diff
@@ -1,26 +1,25 @@
...
   }
 `;
-export class BackendConfigQuery extends Query<{ backendConfig: BackendConfig }> {
+
+export type BackendConfig = {
+  locales: string[];
+  activeLocale: string;
+};
+
+export type BackendConfigQueryResponse = {
```
mateuszwozniak:
```suggestion
export type BackendConfigResponse = {
```

packages/falcon-data/src/BackendConfig/BackendConfigQuery.tsx
```diff
@@ -1,26 +1,25 @@
...
+  backendConfig: BackendConfig;
+};
+
+export class BackendConfigQuery extends Query<BackendConfigQueryResponse> {
```
mateuszwozniak:
```suggestion
export class BackendConfigQuery extends Query<BackendConfigResponse> {
```

packages/falcon-data/src/Url/UrlQuery.ts
```diff
@@ -0,0 +1,43 @@
...
+  type: ResourceType;
+  redirect: boolean;
+  id: string;
+  path: string;
+};
+
+export type UrlQueryVariables = {
+  path: string;
+};
+export type UrlQueryResponse = {
```
mateuszwozniak:
```suggestion
export type UrlResponse = {
```

packages/falcon-data/src/Url/UrlQuery.ts
```diff
@@ -0,0 +1,43 @@
...
+  url: ResourceMeta;
+};
+
+export class UrlQuery extends Query<UrlQueryResponse, UrlQueryVariables> {
```
mateuszwozniak:
```suggestion
export class UrlQuery extends Query<UrlResponse, UrlQueryVariables> {
```

packages/falcon-data/src/BackendConfig/BackendConfigQuery.tsx
tutok: done

packages/falcon-data/src/BackendConfig/BackendConfigQuery.tsx
tutok: done

packages/falcon-data/src/Url/UrlQuery.ts
tutok: done

packages/falcon-data/src/Url/UrlQuery.ts
tutok: done

packages/falcon-front-kit/src/DynamicRoute/DynamicRoute.tsx
```diff
@@ -43,7 +40,9 @@ export const DynamicRoute: React.SFC<DynamicRouteProps> = props => {
 
               const Component = components[url.type];
               if (!Component) {
-                return <Text>{`Please register component for '${url.type}' content type!`}</Text>;
```
Borales: @tutok I would probably keep the message in the body (at least in `development` mode). We can only hope a developer would keep the console tab open during the development, but we cannot guarantee that hence it will confuse new developers with what to do next and why their app does not show anything.

What do you think guys? or maybe `return <NotFound location={location} />;` on line 45?

cc: @mateuszwozniak

packages/falcon-data/src/Url/UrlQuery.ts
```diff
@@ -0,0 +1,43 @@
...
+    url(path: $path) {
+      type
+      redirect
+      id
+      path
+    }
+  }
+`;
+
+export type ResourceType = 'shop-page' | 'shop-product' | 'shop-category' | 'blog-post';
```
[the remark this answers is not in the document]
mateuszwozniak: Good point! Developers can define any kind of custom type and TS would probably complain in that case.

packages/falcon-front-kit/src/DynamicRoute/DynamicRoute.tsx
mateuszwozniak: It might be a good idea for the development mode indeed (just print a warning like it was before).  I don't want to return `NotFound` as that would be misleading - the content have been found but the server doesn't know how to render it. Returning `NotFound` would indicate that passed url cannot be resolved (nobody is able to handle the url) and that's not true in that case.

Borales: @mateuszwozniak so maybe throw an Error and render a nice "error page" in dev mode? (https://github.com/deity-io/falcon/issues/262)

packages/falcon-server/src/schema.graphql
```diff
@@ -63,11 +63,11 @@ type BackendConfig {
   """
   List of supported locales (for example: `["en-US", "nl-NL"]`)
   """
-  locales: [String]
+  locales: [String!]!
```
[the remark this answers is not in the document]
mateuszwozniak: I think when there's only one locale then it should be also listed here to keep things consistent. Otherwise empty list will indicate that there's only one item, and in another place we'll have one item - I don't like that.

packages/falcon-front-kit/src/DynamicRoute/DynamicRoute.tsx
mateuszwozniak: Yes, that would be much better.

packages/falcon-server/src/schema.graphql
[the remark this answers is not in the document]
Borales: closing this comment after offline discussion

packages/falcon-data/src/Url/UrlQuery.ts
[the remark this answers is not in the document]
tutok: done

packages/falcon-data/src/Url/UrlQuery.ts
```diff
@@ -0,0 +1,43 @@
...
+export type UrlResponse = {
+  url: ResourceMeta;
+};
+
+export class UrlQuery extends Query<UrlResponse, UrlQueryVariables> {
+  static defaultProps = {
+    query: GET_URL
+  };
+
+  static propTypes = {
```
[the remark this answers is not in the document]
Borales: closing the comment after offline discussions

packages/falcon-front-kit/src/DynamicRoute/DynamicRoute.tsx
tutok: done

PR #499, opened 2019-06-18
Uikit fix partial typings

mateuszwozniak:

**Other information:**

### DIFF
--- a/packages/falcon-shop-data/src/Address/AddAddressMutation.tsx
+++ b/packages/falcon-shop-data/src/Address/AddAddressMutation.tsx
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag';
 import { Mutation } from 'react-apollo';
 import { AddAddressInput, Address } from '@deity/falcon-shop-extension';
-import { InputArg } from '../types';
+import { OperationInput } from '../types';
 
 export const ADD_ADDRESS = gql`
   mutation AddAddress($input: AddAddressInput!) {
@@ -15,7 +15,7 @@ export type AddAddressResponse = {
   addAddress: Pick<Address, 'id'>;
 };
 
-export class AddAddressMutation extends Mutation<AddAddressResponse, InputArg<AddAddressInput>> {
+export class AddAddressMutation extends Mutation<AddAddressResponse, OperationInput<AddAddressInput>> {
   static defaultProps = {
     mutation: ADD_ADDRESS,
     awaitRefetchQueries: true,
--- a/packages/falcon-shop-data/src/Address/AddAddressMutation.tsx
+++ b/packages/falcon-shop-data/src/Address/AddAddressMutation.tsx
@@ -1,6 +1,7 @@
 import gql from 'graphql-tag';
 import { Mutation } from 'react-apollo';
 import { AddAddressInput, Address } from '@deity/falcon-shop-extension';
+import { InputArg } from '../types';
 
 export const ADD_ADDRESS = gql`
   mutation AddAddress($input: AddAddressInput!) {
@@ -10,15 +11,11 @@ export const ADD_ADDRESS = gql`
   }
 `;
 
-export type AddAddressVariables = {
-  input: AddAddressInput;
-};
-
 export type AddAddressResponse = {
   addAddress: Pick<Address, 'id'>;
 };
 
-export class AddAddressMutation extends Mutation<AddAddressResponse, AddAddressVariables> {
+export class AddAddressMutation extends Mutation<AddAddressResponse, InputArg<AddAddressInput>> {
   static defaultProps = {
     mutation: ADD_ADDRESS,
     awaitRefetchQueries: true,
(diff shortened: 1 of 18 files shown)


packages/falcon-shop-data/src/Address/AddressListQuery.tsx
```diff
@@ -2,14 +2,8 @@ import gql from 'graphql-tag';
...
 
-export type AddressListResponse = {
-  addresses: {
-    items: Address[];
-  };
-};
-
 export const GET_ADDRESS_LIST = gql`
-  query Addresses {
+  query AddressList {
```
Borales: @mateuszwozniak what is the final convention on naming queries/mutations? 

```
mutation AddAddress($input: AddAddressInput!) {
  addAddress ...
}
```

`mutation AddAddress` -> `addAddress` (simple "capitalize")

```
query AddressList {
  addresses { ... }
}
```

`query AddressList` -> `addresses`?

packages/falcon-shop-data/src/Address/AddAddressMutation.tsx
```diff
@@ -8,7 +9,16 @@ export const ADD_ADDRESS = gql`
     }
   }
 `;
-export class AddAddressMutation extends Mutation {
+
+export type AddAddressVariables = {
+  input: AddAddressInput;
```
Borales: @mateuszwozniak since we're shifting into having a single input `input` argument - should we make a generic type for that:

```
// somewhere in "kit"
export type InputArg<T> = {
  input: T;
};

// this file
export class AddAddressMutation extends Mutation<AddAddressResponse, InputArg<AddAddressInput>>
```

or something like that?

packages/falcon-shop-data/src/Address/AddressListQuery.tsx
mateuszwozniak: Right, I missed that one somehow. I'll add a fix shortly.

packages/falcon-shop-data/src/Address/AddAddressMutation.tsx
mateuszwozniak: done

packages/falcon-shop-data/src/Address/AddressListQuery.tsx
mateuszwozniak: done

packages/falcon-shop-data/src/Customer/CustomerWithAddressesQuery.tsx
```diff
@@ -28,7 +28,27 @@ export const GET_CUSTOMER_WITH_ADDRESSES = gql`
   }
 `;
 
-export class CustomerWithAddressesQuery extends Query<Customer> {
+export type CustomerWithAddressesResponse = {
+  customer: Pick<Customer, 'id' | 'firstname' | 'lastname' | 'email'> & {
+    addresses: Pick<
+      Address,
+      | 'company'
```
tutok: Just asking, this `|` should not be after field name?
```suggestion
      'company' |
```

found this: [ts-multipick](https://github.com/staltz/ts-multipick), I am not saying that we should use it but I whant to share it here

packages/falcon-shop-data/src/Customer/CustomerWithAddressesQuery.tsx
```diff
@@ -28,7 +28,27 @@ export const GET_CUSTOMER_WITH_ADDRESSES = gql`
...
+      | 'street'
+      | 'postcode'
+      | 'city'
+      | 'countryId'
+      | 'defaultBilling'
+      | 'defaultShipping'
+      | 'region'
+      | 'regionId'
+      | 'telephone'
+    >;
```
tutok: `customer.addresses` is an array of `Address`
```suggestion
    >[];
```

packages/falcon-shop-data/src/Customer/CustomerWithAddressesQuery.tsx
mateuszwozniak: This actually is autoformatted by our current config, so I'd leave it as it is.

packages/falcon-shop-data/src/types.ts
```diff
@@ -1 +1,4 @@
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+export type InputArg<T> = {
```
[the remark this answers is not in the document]
mateuszwozniak: I was thinking about `Input` but it was too generic in my opinion. `OperationInput` sounds better so I can use it. @Borales are you ok with that? When it comes to moving it - I'm not sure if that makes a sense - atm it's used only in shop related-mutations. We can move it when some other package will need it.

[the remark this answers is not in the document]
Borales: @mateuszwozniak my original comment was addressed to the problem with generic type only, I'm fine with any name you suggest.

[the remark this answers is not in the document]
mateuszwozniak: Cool, renamed to `OperationInput` then.

Issue #411, opened 2019-04-11
ESLint configuration


tutok:
**What should be changed/improved?**
ESLint configuration is too obtrusive for community:
* it is turned on by default (even there is no `.eslintrc` file)
* it is using our configuration 

**Your suggested solution**
Make ESLint optional. Toggling could be achieved via `.eslintrc` file.

tutok:
resolved via #506

PR #506, opened 2019-06-25
ESlint controlled via configuration file existence

tutok:

ES Lint will be turned off by default if the developer would like to turn it on she or he needs to create `.eslintrc` file in the app root
example configuration:

 * `.eslintrc`
```
{
  "extends": ["@deity/eslint-config-falcon"],
  "env": {
    "serviceworker": true
  }
}
```

 * `.eslintignore`
```
build/*
node_modules/*
coverage/*
```

ES Lint was forced by falcon-client

ES Lint is optional

Yes

**Other information:**
No


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Versions marked with a number and date (e.g. Falcon Client v0.1.0 (2018-10-05))
   - upgraded Workbox to the latest 4.3.1 version
   - add build for Service Worker `sw.js` file
 - added source-map support for all modes ([#491](https://github.com/deity-io/falcon/pull/491))
+- improvement: made eslint optional ([#506](https://github.com/deity-io/falcon/pull/506))
```
Borales: @tutok just a suggestion - how about renaming these files with `_disabled` suffix (eg. keeping them as template files without a need to look up for an original content to be able to enable them), so if a developer would need to enable eslint - simply renaming these files back to normal names would suffice? 

`.eslintignore_disabled` and `.eslintrc_disabled`

cc @mateuszwozniak

tutok: it is not a that bad idea but... then I think we need to do the same thing with `.babelrc` for example.
For sure I will update docs so it will be explained what needs to be done to configure ESLint.

Borales: 👍 sure

or simply replace `.` with a `_` (so they would stay on top of the list + no need to change the original filename)

mateuszwozniak: If we have docs then we don't need suffixes etc. @tutok please add also info what's in default .eslintrc (import of falcon eslint preset).

Borales: if we're removing these files - in this case, the CHANGELOG line has to be changed to "disabling eslint by default" (it's not optional anymore)

tutok: One important note. My initial idea was to solve this https://github.com/deity-io/falcon/issues/411, especially the first bullet:

 > *it is turned on by default (even there is no .eslintrc file)*
 
Why? Because we introducing more and more strict linting rules (with upcoming release too) because we like it, but we receive also the clear message from the community that they want to have the option to throw away linting

In this PR I did not want to remove `.eslintrc` at all nor change its name in order to disable config, to be honest. But introduce only possibility to *turn off* linting. So, the best solution was to check if `.eslintrc` file exists, if yes then incorporate `eslintLoader`, if not then do nothing. I wanted to leave `.eslintrc` in example app so no breaking changes.

Right after an above discussion about "how to remove `.eslintrc` from example project" I started thinking about the problem which you guys want to solve (which is different I think). And I think we could provide separate `eslint` presset with syntax checks only without code formatting, without prettier, allowing duplicated imports and so on and so on. Personally, I would not like to remove `.eslintrc` files (nor provide templates) because we can provide default config which is more suited for our example app.

What I suggest is:
- merge possibility of turning off eslint via removing `.eslintrc` file
- leave `.eslintrc` and `.eslintignore` files as it was before.
- create another task to prepare less obtrusive (syntax only) eslint preset, and latter use it instead of current `@deity/eslint-config-falcon`.


What do you think? 
If you do not like the above suggestion please tell, then :
- I will merge this PR (so `.eslintrc` and `.eslintignore` will be removed)
- I will add also appropriate documentation update

Borales: @tutok I like your suggestions, let's do it this way

PR #507, opened 2019-06-26
Uikit rework - mini cart and related components

mateuszwozniak:
Parts moved in this pull request:
- Queries and mutations related to cart operations
- MiniCart, MiniCartProduct and their layouts
- Price component
- DateFormat (renamed to FormattedDate)
- Updated typings to include `__typename` field


packages/falcon-shop-data/src/Cart/AddToCartMutation.tsx
```diff
@@ -0,0 +1,28 @@
...
+`;
+
+export type AddToCartResponse = {
+  addToCart: CartItemPayload;
+};
+
+export class AddToCartMutation extends Mutation<AddToCartResponse, { input: AddToCartInput }> {
+  static defaultProps = {
+    mutation: ADD_TO_CART,
+    awaitRefetchQueries: true,
```
Borales: @mateuszwozniak do we have mutations where this flag is set to `false`? Can we change the default behavior through the ApolloClient settings instead of passing the same flag to every Mutation we have? (https://www.apollographql.com/docs/react/api/apollo-client/#apolloclient - `defaultOptions.mutate`)

It would be still possible to disable `awaitRefetchQueries` by passing `false` value, but by default - it's gonna be enabled.

packages/falcon-shop-data/src/Cart/RemoveCartItemMutation.tsx
```diff
@@ -0,0 +1,44 @@
...
+export type RemoveCartItemResponse = {
+  removeCartItem: RemoveCartItemPayload;
+};
+
+export class RemoveCartItemMutation extends Mutation<RemoveCartItemResponse, { input: RemoveCartItemInput }> {
+  static defaultProps = {
+    mutation: REMOVE_CART_ITEM,
+    refetchQueries: ['Cart'],
+    awaitRefetchQueries: true,
+    update: (
```
Borales: @mateuszwozniak just a crazy idea/suggestion. How about extracting such `update` callbacks into separate vars (like we do for GQL queries - if somebody would want to use them in their Mutation components) or even to separate files (to be able to alter/replace those `update` callbacks via `normal-module-replacement`)?

It would also make our exported classes a bit cleaner

cc @tutok

packages/falcon-shop-data/src/Cart/AddToCartMutation.tsx
mateuszwozniak: Yes, you're right - we have only one place where it's not used and in fact it's probably just missed. I'll take care of that.

packages/falcon-client/src/clientApp/defaultConfiguration.js
```diff
@@ -16,7 +16,11 @@ export default initialConfig =>
         },
         connectToDevTools: process.env.NODE_ENV !== 'production',
         defaultOptions: {
-          __typename: 'ApolloClientDefaultOptions'
+          __typename: 'ApolloClientDefaultOptions',
+          mutate: {
+            __typename: 'ApolloClientDefaultMutateOptions',
+            awaitRefetchQueries: true
```
Borales: @mateuszwozniak just for the sake of keeping falcon-client config "clean" - maybe we could better pass this config directly in `packages/falcon-client/src/service/ApolloClient.js:66`? apart from that - this flag is pretty crucial for our UI/components, and if developers would miss this flag during the migration to the newer version - something may go wrong.

packages/falcon-shop-data/src/Cart/AddToCartMutation.tsx
Borales: @mateuszwozniak just for the sake of keeping falcon-client config "clean" - maybe we could better pass this config directly in `packages/falcon-client/src/service/ApolloClient.js:66`? apart from that - this flag is pretty crucial for our UI/components, and if developers would miss this flag during the migration to the newer version - something may go wrong.

mateuszwozniak: No problem, that makes sense. I'll move it.

packages/falcon-shop-data/src/Cart/AddToCartMutation.tsx
```diff
@@ -0,0 +1,28 @@
...
+      productType
+    }
+  }
+`;
+
+export type AddToCartResponse = {
+  addToCart: CartItemPayload;
+};
+
+export class AddToCartMutation extends Mutation<AddToCartResponse, { input: AddToCartInput }> {
```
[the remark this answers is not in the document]
mateuszwozniak: done

packages/falcon-shop-data/src/Cart/AddToCartMutation.tsx
mateuszwozniak: done

packages/falcon-shop-extension/src/schema.graphql
```diff
@@ -434,15 +434,15 @@ type CartItem {
 }
 
 type CartItemOption {
-  label: String
-  value: String
-  data: [CartItemOptionValue]
+  label: String!
+  value: String!
+  data: [CartItemOptionValue]!
```
tutok: I think we want to have a not nullable collection of items, am I wrong?
```suggestion
  data: [CartItemOptionValue!]!
```

packages/falcon-shop-extension/src/schema.graphql
```diff
@@ -28,7 +28,7 @@ extend type Mutation {
   setShopStore(storeCode: String!): BackendConfig
   addToCart(input: AddToCartInput!): CartItemPayload
   updateCartItem(input: UpdateCartItemInput!): CartItemPayload
-  removeCartItem(input: RemoveCartItemInput!): RemoveCartItemResponse
+  removeCartItem(input: RemoveCartItemInput!): RemoveCartItemPayload
   applyCoupon(input: CouponInput!): Boolean
   cancelCoupon: Boolean
```
tutok: I think we could have only two states (`true` or `false`) here?
```suggestion
   cancelCoupon: Boolean!
```

tutok:
thanks

PR #520, opened 2019-07-09
Auto-installing Subscription handlers

Borales:


Borales:
@mateuszwozniak please review this PR once again (added handling cookie/session and headers data for Subscriptions to ensure `context` is being properly initialized)

packages/falcon-server/src/index.js
```diff
@@ -316,18 +339,72 @@ class FalconServer {
...
+  /**
+   * Get default Subscriptions Options with even handlers to properly initialize required context values
+   */
+  getSubscriptionsOptions() {
+    return {
+      onConnect: (connectionParams, websocket, context) => {
+        const { cookie: paramsCookie } = connectionParams;
+        if (paramsCookie) {
+          // Forwarding `cookie` from `connectionParams` to the request object
+          context.request.headers.cookie = paramsCookie;
```
tutok: you are actually overwriting existing `request.headers.cookies` via `paramsCookie` instead of forwarding them. Moreover, changing the entire `context.request.headers` because of threating it as a local variable is considered as a code smell.

Borales: since there were no suggestions, I changed the code the preserve the original `context.request` and pass the "fake" one to the Cookie parser.

Issue #426, opened 2019-04-26
[Filters] Category Filter returns no results


cane-lagerwaard:
**Describe the bug**
When selecting a category within the Filter options, no results are found

**To Reproduce**
1. Go to https://latest.snapshot.deity.io/men.html
2. Select a category filter, e.g. `Tops`
3. Notice 0 products displayed/found

**Expected behavior**
Either Category filter should not be visible, or it should display results. See here: https://latest.snapshot.deity.io/men.html?filters[cat]=12

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5bc9eac631854f160a8b5fe8/c48e9a46-ce4f-4129-8cc4-c1fd407609b5)

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser Chrome

tutok:
**Your suggested solution**
I think it is due to the wrong implementation both UI and Magento2Api extension.

* UI part - I think we should "reload page" when another category is selected because we should show "category" related stuff (filters, subcategories, products).

* Magento2Api part - currently we show products for the parent category `X` even we select its child category `Y` so that is why we can not see the proper result. When subcategory `Y` is selected, then category `Y` should be fetched with its products. There is also a known issue that we do not have access to child categories URL path, we have only its id's. We can overcome this via `/categories/list` call which will return all categories (with its properties like URL path), so later we will be able to map category id to category URL Path.

PR #524, opened 2019-07-18
Use category's id from filters list when querying

akbarik:

It fixes the bug described in this [issue](https://github.com/deity-io/falcon/issues/426).

If you select Men or Women from the Header and then click on any filter in Category list, you will get 0 results, even though in filters it shows that it has products.

### DIFF
--- a/examples/shop-with-blog/client/src/pages/shop/Category/Category.js
+++ b/examples/shop-with-blog/client/src/pages/shop/Category/Category.js
@@ -23,14 +23,24 @@ import { Filters } from './Filters';
 
 const copy = item => item && JSON.parse(JSON.stringify(item));
 
+// if product's category is in filters, use that as a value for categoryId parameter of CategoryProductsQuery
+const getId = (id, filters) => {
+  let resultId = id;
+  const category = filters.find(item => item.field === 'cat');
+  if (category && category.value) {
+    resultId = Array.isArray(category.value) && category.value.length > 0 ? category.value[0] : category.value;
+  }
+  return resultId;
+};
+
 const CategoryPage = ({ id }) => (
   <SearchConsumer>
     {({ state }) => (
       <CategoryProductsQuery
         variables={{
-          categoryId: id,
+          categoryId: getId(id, state.filters),
           sort: state.sort,
-          filters: copy(state.filters)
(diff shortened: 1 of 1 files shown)


Borales:
In my opinion, we should be dealing with this bug slightly different.
In the suggested PR - we're trying to fix a Magento's bug on the frontend when it cannot apply the provided filters to the query (meaning - fixing a backend-specific issue on the "agnostic" frontend).

I would suggest one of the following solutions:
- exclude `cat` filter in the `magento2-api` package response at all as we have these categories in Men/Women top sub-menus
- we should render `categories` filter block as links, meaning - no filter processing and simply navigate  to the specified sub-category

@cane-lagerwaard @mateuszwozniak @tutok @akbarik what do you guys think?

akbarik:
> * exclude `cat` filter in the `magento2-api` package response at all as we have these categories in Men/Women top sub-menus
No, actually there are more categories that you can recieve using `cat` filter. Try to click on Men/Women -> Tops in the menu. Then expand Category filter group, you'll see several categories there, like Jackets, Tees and so on. They don't exist in the menu.

Borales:
@akbarik sorry for a delayed response.

My point was about fixing this behavior on the `magento2-api` package itself. Apparently, Magento is unable to apply its own filters 🤷‍♂ (selecting any "Category" filter on https://demo.deity.io/men.html shows "0 products"). But if you check a sub-category page (https://demo.deity.io/men/tops-men.html) - now you can select them. I think it has something to do with filtering by category on a "leaf" category (correct me if I'm wrong @sergy-gzh ). 

To summarize: I'd suggest leaving frontend as is, and try to adapt `magento2-api` package to work properly with the filters

Borales:
@akbarik 

> Then expand Category filter group, you'll see several categories there, like Jackets, Tees and so on. They don't exist in the menu.

This is a matter of a frontend template, by default, it renders 2 menu levels only, but the menu could have a lot more levels.

akbarik:
Ok, I see what you mean, @Borales. Agreed, it has to be fixed on magento2-api level. Closing this PR.

Borales:
@akbarik just in case if you'll be working on this issue - https://github.com/deity-io/falcon/issues/426#issuecomment-527918814

PR #540, opened 2019-07-30
fix: checkout address form submission failed

nimobeeren:

Bug fix.

Submitting the address form in checkout would cause an error: `Network error: Response not successful: Received status code 400`. This was introduced in #535, where some form field IDs were changed.

The IDs were reverted, fixing the error and letting the user continue.

No.


nimobeeren:
thanks :+1:

PR #549, opened 2019-08-09
Adding TTL for cache tags

Borales:

Bug fix

When using Redis/Memcached as a cache backend and setting value without a TTL option - Apollo forces `300` seconds (5 min) TTL by default (meaning - there's no way to set a cached value without an expiration):

- https://github.com/apollographql/apollo-server/blob/master/packages/apollo-server-cache-redis/src/RedisCache.ts#L8
- https://github.com/apollographql/apollo-server/blob/master/packages/apollo-server-cache-memcached/src/index.ts#L9

Setting 1 hour TTL for cache tags.

No

### DIFF
--- a/packages/falcon-server-env/src/cache/Cache.ts
+++ b/packages/falcon-server-env/src/cache/Cache.ts
@@ -26,8 +26,7 @@ export type GetCacheOptions = {
   options?: SetCacheOptions;
 };
 
-// 1 hour
-const DEFAULT_TAG_TTL: number = 60 * 60;
+const DEFAULT_TAG_TTL: number = 60 * 60; // 1 hour
 
 /**
  * Cache-wrapper with extended methods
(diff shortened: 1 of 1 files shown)


packages/falcon-server-env/src/cache/Cache.ts
```diff
@@ -26,11 +26,14 @@ export type GetCacheOptions = {
   options?: SetCacheOptions;
 };
 
+// 1 hour
+const DEFAULT_TAG_TTL: number = 60 * 60;
```
tutok: I am aware that it could be considered fussy but I would put comment from the line `29` into `30`
```suggestion
const DEFAULT_TAG_TTL: number = 60 * 60; // 1 hour
```

Borales: sure, done

Issue #425, opened 2019-04-26
[Filters] Selecting Price Filter does not display correct results


cane-lagerwaard:
**Describe the bug**
When selecting a price on a category page, it doesn't filter the results for that price range

**To Reproduce**
1. Go to https://latest.snapshot.deity.io/men.html
2. Select a price filter
3. Notice results don't change

**Expected behavior**
Results should be adjusted to the used filters

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5bc9eac631854f160a8b5fe8/d76bf18d-c386-4978-87e3-1f4ec6f14cb2)

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser Chrome

tutok:
it is probably connected with #417, because special prices and tier prices are taken into account same like on sorting

Borales:
@cane-lagerwaard can we close this? seems like it was fixed

cane-lagerwaard:
@Borales this issue is still there for me

PR #562, opened 2019-08-27
fix: #425 Price range filter

strrife:

- [+] The commit message follows our guidelines

Fix for https://github.com/deity-io/falcon/issues/425

Now the price filter does not work at all

Price filter works, except for cases described in the issue I am creating right now

Basically the problem is with the app itself rather than packages, so to fix this in an already existing store one would have to fix the filter logic.


tutok:
@strrife thanks for your contribution!

PR #570, opened 2019-09-03
Update Schema for address input

philkun:
Update Schema for address input to allow capturing additional data like with vanilla magento customer addresses

**Other information:**

### DIFF
--- a/packages/falcon-shop-extension/src/schema.graphql
+++ b/packages/falcon-shop-extension/src/schema.graphql
@@ -199,6 +199,7 @@ input AddAddressInput {
   countryId: String!
   defaultBilling: Boolean
   defaultShipping: Boolean
+  region: String
   regionId: Int
 }
 
@@ -214,6 +215,7 @@ input EditAddressInput {
   countryId: String!
   defaultBilling: Boolean
   defaultShipping: Boolean
+  region: String
   regionId: Int
 }
 


Borales:
Hi @philkun
Thank you for your PR. Does it require any additional changes to magento2-api package?

packages/falcon-shop-extension/src/schema.graphql
```diff
@@ -214,6 +215,7 @@ input EditAddressInput {
   countryId: String!
   defaultBilling: Boolean
   defaultShipping: Boolean
+  region: String
```
tutok: In my opinion, the `region` name should not be exposed in that way.

We do not have `country` field on Address, so why we try to introduce the `region`?  I think we should be consistent here, having `regionId` should be enough as `countryId` is. If someone would like to display the region name, he should take it from "dictionary", so we should introduce a possibility to fetch all regions for the specific `country` instead.

Eventually, we could:
`countryId: string!` --> `country: Country!` and 
`regionId: string` --> `region: Region`

philkun: This is a Magento thing, so I am not sure if an alternative  implementation needs to be looked into.  Magento region_id is not known for countries unless you force every country to have a state, exposing the region field gives the ability and flexibility to replicate Magento functionality where it is optional as a text input e.g. 
See short animated gif as an example
https://i.gyazo.com/c6dd5a2be634eed6df3a05099c85ead6.gif

If you visit the demostore, add to basket and  get to checkout, select france as a country, you should see the state auto populated(region_id), selecting a country like UK  where the state is not required reveals a text input (region) 

https://magento.deity.io/cassius-sparring-tank.html

philkun: Also no  code changes are necessary on the magento2-api package as this is just exposing the field region.

nimobeeren:
Hi @philkun!
In my PR #609 I changed the field `region: Int` to `region: ID`. For an `ID` type, you can pass either a `String` or an `Int`, so it is up to the back-end to decide how they want to identify a region.
I would love to hear your thoughts on this.

nimobeeren:
Hello @philkun,
Since #607 has been merged, could you take another look to see if this helps you do what you want?

While the `AddressInput` still requires a `regionId`, the `Address` type now contains more details about the region.

Feel free to re-open this if you believe something can be improved.

Issue #578, opened 2019-09-05
Move sortOrders from shop-with-blog into Magento2-api


tutok:
**What should be changed/improved?**
sort orders (the default items collection) should be returned via shop-extension implementation (via config resolver) instead of defined in shop-with-blog example app. Because falcon-front-kit uses them

PR #591, opened 2019-09-13
Moved sortOrders into Magento2-api

tutok:

resolved https://github.com/deity-io/falcon/issues/578

yes

**Other information:**
no


packages/falcon-server/schema.graphql
```diff
@@ -128,6 +128,22 @@ input SortOrderInput {
...
 
+type SortOrderValue {
+  """
+  Perform sorting by this field
+  """
+  field: String!
+  """
+  Sort order direction
+  """
+  direction: SortOrderDirection
```
Borales:
```suggestion
  direction: [SortOrderDirection!]
```

@mateuszwozniak what do you think of making this as `direction: [SortOrderDirection!]`?

The response is gonna look like:

```
name: "Price",
direction: [{
  field: 'product_price',
  direction: ASC
}, {
  field: 'product_price',
  direction: DESC  
}]
```

Right?

Otherwise, we could flatten this structure and get off this unnecessary nesting:


```graphql
type SortOrder {
  name: String!
  field: String!
  value: SortOrderValue
}
```

mateuszwozniak: I don't think we'll need sorting by more than one criteria so I'd rather flatten that instead of using array.

PR #594, opened 2019-09-19
Fix ServiceWorker redirects

### LINKED ISSUE #593: ServiceWorker and protected routes
**Describe the bug**
SW is unable to handle redirects for protected routes on the client-side.

**To Reproduce**
1. Go to https://demo.deity.io/account/orders/1
2. See the error

```
The FetchEvent for "https://demo.deity.io/account/orders/1" resulted in a network error response: a redirected response was used for a request whose redirect mode is not "follow".
```

nimobeeren:

Bug fix.

Fixes #593.

Although the redirect works, a new error is printed to console:

`index.js:1 Warning: Prop `href` did not match. Server: "/sign-in" Client: "/account/orders/1"`

Not sure how to fix that.

No.

**Other information:**
Based on https://stackoverflow.com/questions/45434470/only-in-chrome-service-worker-a-redirected-response-was-used-for-a-reque.


packages/falcon-client/src/serviceWorker/sw.js
```diff
@@ -39,10 +39,26 @@ if (CONFIG.precache) {
...
+  });
+}
+
 const router = new Router();
 self.addEventListener('fetch', event => {
-  const responsePromise = router.handleRequest(event);
+  let responsePromise = router.handleRequest(event);
   if (responsePromise) {
+    responsePromise = responsePromise.then(res => (res.redirected ? cleanResponse(res) : res));
     event.respondWith(responsePromise);
```
tutok: can not we do 
```suggestion
     event.respondWith(responsePromise.then(res => (res.redirected ? cleanResponse(res) : res)));
```
?
then `responsePromise` can be still const.

The second thing is why we need to do "response to blob rendering" only when there is redirect? should not we do it for all possible responses?

packages/falcon-client/src/serviceWorker/sw.js
```diff
@@ -39,10 +39,26 @@ if (CONFIG.precache) {
   precacheAndRoute(ENTRIES, {});
 }
 
+async function cleanResponse(response) {
```
tutok: I am not sure if the name of this function describes what is doing

Borales:
@nimobeeren should we consider including "external redirects" into this PR also? 

Like `<Redirect to="http://example.com" />`

_Although, it also could be considered as a separate task._

nimobeeren:
> @nimobeeren should we consider including "external redirects" into this PR also?

@Borales are those also broken then?

Borales:
@nimobeeren looks like so, I was testing such redirection, it was partially working, and when it was working - I got redirected to `http://localhost:3000/http://example.com`

nimobeeren:
@Borales I can take a look at that within this PR. Marking this as WIP

packages/falcon-client/src/serviceWorker/sw.js
```diff
@@ -43,7 +43,9 @@ const router = new Router();
 self.addEventListener('fetch', event => {
   const responsePromise = router.handleRequest(event);
   if (responsePromise) {
-    event.respondWith(responsePromise);
+    // if response is redirected, we should clone the response before sending it
+    // see https://stackoverflow.com/a/45440505/412319
+    event.respondWith(responsePromise.then(res => (res.redirected ? res.clone() : res)));
```
tutok: @nimobeeren why the previous `cleanResponse` function was removed?

```suggestion
    event.respondWith(responsePromise.then(res => (res.redirected ? cleanResponse(res) : res)));
```
I did a couple of tests, and the current implementation does not work on my side. I am receiving the same error as in reported issue.

Alo did som tests with suggested SO solution and it looks fine. Maybe we misunderstood in one of the previous comments?

mateuszwozniak: I also did some tests and it looks like @tutok is right - the solution with `cleanResponse(res)` works correctly for me too, while version from this PR doesn't so @nimobeeren please update it.

PR #596, opened 2019-09-20
Update to react apollo 3.1.x and apollo client 2.6.x (hooks support)

### LINKED ISSUE #597: Confusing resolutions entry in example
**What should be changed/improved?**
The current `resolutions` field int he example sets the `apollo-env` to `0.3.1` but package.json says / wants `0.4.0`.

**Your suggested solution**
Should the resolutions filed be removed or updated to `"apollo-env": "0.3.1"`?

https://github.com/deity-io/falcon/blob/master/examples/shop-with-blog/server/package.json#L32

mateuszwozniak:
First approach to the update of react-apollo so hooks can be used. 

Still requires a bunch of tweaks to apollo packages and imports but `Home.js` already uses `useQuery` hook with our query to fetch products for homepage.

It also looks that SSR works correctly but it also requires a bit more testing as I've just done superficial testing.

Fixes #597


packages/falcon-blog-data/package.json
```diff
@@ -31,16 +31,16 @@
...
+    "apollo-client": "2.6.4",
     "graphql-tag": "2.10.1",
     "prop-types": "15.7.2",
     "react": "16.8.6",
-    "react-apollo": "2.5.8"
+    "react-apollo": "3.1.1"
   },
   "peerDependencies": {
-    "apollo-client": "2.5.1",
+    "apollo-client": "2.6.4",
```
Borales: @mateuszwozniak no `^` support for this package? or it's too risky?

packages/falcon-data/src/BackendConfig/SetLocaleMutation.tsx
```diff
@@ -1,5 +1,6 @@
 import gql from 'graphql-tag';
-import { Mutation } from 'react-apollo';
+// import { Mutation } from '@deity/falcon-data';
```
Borales:
```suggestion
```

mateuszwozniak:
@nimobeeren I wanted to avoid that, but Apollo changed the implementation of `Query` and `Mutation` from classes that extend `React.Component` to pure functions, so we couldn't use these as base classes for our Queries/Mutations. 
Now, our base Query/Mutation classes just wrap Apollo's and we can deliver all the components without changes in the implementation of our already implemented Query/Mutation components.

nimobeeren:
I see, seems like they are going functional/hooks all the way

packages/falcon-blog-data/package.json
mateuszwozniak: done

PR #609, opened 2019-09-26
Schema change: address/country/region

nimobeeren:

Schema update and new features in `falcon-magento2-api`.

* `type Address` now contains fields `country: Country` and `region: Region`, which allows querying more data than just the country/region ID.
* `type Country` now contains both `id: ID!` and `code: String`, which is useful for back-ends that make a distinction between the two.

Yes. Instead of querying the field `Address.countryId` or `Address.region` to get the country/region ID, you should now query `Address.country.id` or `Address.reigon.id`, respectively.

> Will wait for PR #638 before continuing


nimobeeren:
@Borales I think that's a good idea, it will also help clear up any confusion about the `countryId` field (sometimes it's a numerical ID, other times a code made of letters).

I mentioned before that I wanted to merge all types related to `Address` together, but combined with your suggested change we have the issue that for an `AddressInput` we cannot expect the user to provide a complete `Country` object (with all its subfields). So in that case it might be better to keep `Address` and `AddressInput` as separate types.

Borales:
@nimobeeren I only mentioned `type X` statements. There's no need to pass the whole `Country` or `Region` object as a part of the `input` value, we only need an `ID` value to be passed through to the backend (it could be an `int` or a `string` - depending on what the backend sets as a `Country.id` value).

`Address.country` type mainly refers to the problem with nested data and UI.

packages/falcon-magento2-api/src/index.js
```diff
@@ -1240,6 +1240,34 @@ module.exports = class Magento2Api extends Magento2ApiBase {
...
   }
 
+  /**
+   * Get country details
+   * @param {object} obj parent object
+   * @param {ID} obj.countryId country ID
+   * @returns {promise<Country>} requested country data
+   */
+  async country({ countryId }) {
+    const countries = await this.countryList();
```
tutok: I think we should also update `this.countryList()` resolver to cache mapped response:

```
  /**
   * Fetch country list
   * @returns {CountryList} parsed country list
   */
  async countryList() {
    const countries = await this.get(
      '/directory/countries',
      {},
      {
        context: {
          didReceiveResult: result =>
            result.map(item => ({
              code: item.id,
              englishName: item.full_name_english,
              localName: item.full_name_locale,
              regions: item.available_regions || []
            }))
        }
      }
    );

    return { items: countries };
  }
```

packages/falcon-shop-extension/schema.graphql
```diff
@@ -33,7 +33,7 @@ extend type Mutation {
   resetPassword(input: ResetPasswordInput!): Boolean
   editAddress(input: EditAddressInput!): Address
   addAddress(input: AddAddressInput!): Address
-  removeAddress(id: Int!): Boolean
+  removeAddress(id: ID!): Boolean
```
tutok:
```suggestion
  removeAddress(id: ID!): Boolean!
```

packages/falcon-shop-extension/schema.graphql
```diff
@@ -533,9 +533,8 @@ type Address {
   street: [String!]!
   city: String!
   postcode: String!
-  countryId: String!
-  region: String
-  regionId: Int
+  country: Country!
```
tutok: I am just courios if we are aware that `Country` type actually contains property `regions`. So, in the end, it will be possible to fetch 
```
  address {
    country {
       regions {
         id
       }
    }
    region {
      id
    }
  }
```

maybe we should not allow doing this here, and use here `CountryWithoutRegions` type (type name roughly invented).

Still, we should use `countryList` to get `Country`ies  with `regions`

packages/falcon-magento2-api/src/index.js
```diff
@@ -65,6 +64,10 @@ module.exports = class Magento2Api extends Magento2ApiBase {
       },
       PaymentMethod: {
         config: apiGetter((api, ...args) => api.getPaymentMethodConfig(...args))
+      },
+      Address: {
+        country: apiGetter((api, ...args) => api.country(...args)),
+        region: apiGetter((api, ...args) => api.region(...args))
```
tutok: I would rename it to `getAddressRegion` as this resolver is `Address` specific (we have a similar situation with `Product` and `this.productPrice`)

```suggestion
        region: apiGetter((api, ...args) => api.addressRegion(...args))
```

packages/falcon-shop-extension/schema.graphql
nimobeeren: Actually, I don't really see a reason why we wouldn't allow this way of getting regions through country. We replaced the `countryId: ID` field with `country: Country` so the client can get more details of the country, so why not the regions as well?

packages/falcon-magento2-api/src/index.js
nimobeeren: Done :+1:

packages/falcon-magento2-api/src/index.js
```diff
@@ -897,13 +900,23 @@ module.exports = class Magento2Api extends Magento2ApiBase {
...
-    const countries = response.map(item => ({
-      code: item.id,
-      englishName: item.full_name_english,
-      localName: item.full_name_locale,
-      regions: item.available_regions || []
-    }));
+    const countries = await this.getAuth(
+      '/directory/countries',
+      {},
+      {
```
Borales:
```suggestion
      {
        cacheOptions: { ttl: 86400 }, // 24 hours cache
```

I think it would be safe to cache that for 24 hours (doing this on API class, because other methods are also using this getter)

packages/falcon-shop-extension/schema.graphql
Borales: I agree with @nimobeeren here. If for some reason a developer would need to get a list of available regions for this specific country on the address page - he can do it (as long as it's a valid GQL request). I'd keep GQL schema with Nimo's changes as is, this is the core idea of GraphQL. 

Another tiny issue I found in the schema - `Query.countryList` returns `CountryList!` with the following definition:

```graphql
type CountryList {
  items: [Country!]!
}
```

Do we really need this nested value? Can't we do simply `Query { countryList: [Country!]! }`?

nimobeeren: @Borales in this case I think it makes sense to leave it as is, for the sake of consistency with the other `List` types. Some of them have additional fields like `pagination` and `aggregation` alongside `items`.

packages/falcon-magento2-api/src/index.js
nimobeeren: Do you know what is the default TTL?

Borales: @nimobeeren there's no default TTL (it works automatically if a backend sends `max-age` response header - this way TTL is set from that value). Otherwise - there's no default value.
So sometimes, we need to force this cache TTL. _It's just a suggestion_

CHANGELOG.md
```diff
@@ -62,6 +62,11 @@ Versions marked with a number and date (e.g. Falcon Client v0.1.0 (2018-10-05))
 - initial release
 - added Apollo hooks support (upgrade `apollo-client` to version `2.6.x`) ([#596](https://github.com/deity-io/falcon/pull/596))
 
+### Falcon Shop Extension v1.0.0
+
+- initial release
```
Borales:
```suggestion
```

that's not an initial release

packages/falcon-shop-extension/schema.graphql
Borales: @nimobeeren sure, I don't mind.

@tutok it's up to you to decide whether we should close this thread, I got my answers.

packages/falcon-magento2-api/src/index.js
nimobeeren: So when a backend doesn't send such a header, is the value cached indefinitely or not at all? I just want to know how it works :)

Borales: @nimobeeren it works for positive values only, so if there's no `max-age` header - nothing will be cached (meaning - refetch from the source each time).

packages/falcon-ui-kit/src/Address/AddressDetails.tsx
```diff
@@ -9,7 +10,8 @@ export type AddressDetailsProps = {
   street: string[];
   postcode?: string;
   city: string;
-  countryId: string;
+  country?: Country;
+  countryId?: string;
```
Borales: @nimobeeren `countryId` covers the Checkout address? (just a question)

packages/falcon-front-kit/src/Address/addressToString.ts
```diff
@@ -5,7 +5,7 @@ export const addressToString = (address: Address) =>
     address.company,
     `${address.firstname} ${address.lastname}`,
     address.street && `${address.street.join(' ')}`,
-    `${address.postcode} ${address.city}, ${address.countryId}`
+    `${address.postcode} ${address.city}, ${address.country.code}`
```
tutok: base on out graphql schema can we assume that `code` is available here? I think no, and I think we should update schema 😃

packages/falcon-shop-data/src/Address/AddressListQuery.tsx
```diff
@@ -35,8 +39,8 @@ export type AddressListResponse = {
       | 'city'
       | 'postcode'
       | 'region'
-      | 'regionId'
-      | 'countryId'
+      | 'region'
+      | 'country'
```
tutok: I think you should update AddressListResponse type, in following way:
```
export type AddressListResponse = {
  addressList: {
    items: (Pick<
      Address,
      | 'id'
      | 'firstname'
      | 'lastname'
      | 'telephone'
      | 'street'
      | 'city'
      | 'postcode'
      | 'company'
      | 'defaultBilling'
      | 'defaultShipping'
    > & {
      region: Pick<Region, 'id'>;
      country: Pick<Country, 'id' | 'code'>;
    })[];
  };
};
```

`Pick` works only on one level, while we do not select everything for `region` and `country` here,

the same thing for `AddressQuery` and `CustomerWithAddressesQuery` and `OrderQuery`

packages/falcon-shop-extension/schema.graphql
```diff
@@ -51,7 +51,7 @@ extend type Mutation {
   resetPassword(input: ResetPasswordInput!): Boolean
   editAddress(input: EditAddressInput!): Address
   addAddress(input: AddAddressInput!): Address
-  removeAddress(id: Int!): Boolean
+  removeAddress(id: ID!): Boolean!
```
tutok: some time ago I have a discussion with @Borales about `Boolean` result for mutations which should return `void` and I think in that situation we should use `Boolean`
```suggestion
  removeAddress(id: ID!): Boolean
```

as it is not possible to return `void`. Why `void`? I think if `removeAddress` can not succeed we should throw error.
ref: https://blog.apollographql.com/full-stack-error-handling-with-graphql-apollo-5c12da407210

Borales: I agree with @tutok , returning an optional `Boolean` seems like a "lesser evil" for us (since GQL mutation needs to have a returning type defined in the Schema). 

By returning nothing from the mutation resolver - GQL shows it as `null` in the response (which is kind ok for us, because in our Mutation responses we don't check the returning value, the only "negative" result will be returned in a form of error).

packages/falcon-shop-extension/src/types.ts
```diff
@@ -369,16 +372,17 @@ export type CountryList = {
 };
 
 export type Country = {
-  code: string;
-  englishName: string;
-  localName: string;
-  regions: Region[];
+  id: number;
+  code?: string;
```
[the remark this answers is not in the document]
nimobeeren: done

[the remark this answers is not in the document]
nimobeeren: done

packages/falcon-shop-extension/schema.graphql
```diff
@@ -668,15 +667,24 @@ type BundleProductOptionLink {
...
+  # region name
   name: String
 }
 
 type Country {
-  code: String!
+  # unique identifier of the country
+  id: ID!
+  # abbreviated form of the country name
+  code: String
```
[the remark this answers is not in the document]
nimobeeren: done

packages/falcon-front-kit/src/Address/addressToString.ts
nimobeeren: As discussed on Slack, we will change `Country.code` to be required in schema, because the UI depends on it.

packages/falcon-shop-extension/schema.graphql
nimobeeren: I agree with you guys, the mutation result type should not be optional *just* because it may be `null` when an error occurs. If were to follow that, then every return type should be optional. The return type should indicate what to expect under normal conditions (no errors), hence it should be required.

nimobeeren: I initially misread your comment, here is my updated opinion:

I disagree on making the mutation result type optional. It should not be optional *just* because we may respond with `null` when an error occurs. If were to follow that, then every return type should be optional (because an error can always occur). The return type should indicate what to expect under normal conditions (no errors), hence it should be required.

packages/falcon-shop-data/src/Address/AddressListQuery.tsx
nimobeeren: done

examples/shop-with-blog/client/src/pages/shop/Checkout/AddressSection.js
```diff
@@ -25,9 +25,17 @@ class AddressSection extends React.Component {
...
     // make sure we don't send __typename field
     const addressInput = { ...selectedAddress };
     delete addressInput.__typename; // eslint-disable-line no-underscore-dangle
+
+    // include ID fields instead of full country/region objects
+    addressInput.regionId = selectedAddress.region ? selectedAddress.region.id : undefined;
+    delete addressInput.region;
+    addressInput.countryId = selectedAddress.country.id;
+    delete addressInput.country;
+
```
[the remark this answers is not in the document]
nimobeeren: I like that. I also included `__typename` in that destructuring assignment to take care of L29-31.

packages/falcon-shop-extension/schema.graphql
tutok: we do not want to say that return value should be always optional for mutation. When we add new Address we should return a newly added entity, so the return type should be `Address!`. But here we talk about removing the entity, so I think `void` (`Boolean` in our case) will be perfect here.

I think we should be consistent in returning errors, to be able to build on top of this consistency the same error handling mechanism.
If we can throw an error if a user input for the query is not correct then why we can not throw an error if we can not successfully remove the entity?

Side note, when you throw an error inside a function body then the function will not return any result, even `void` :)

packages/falcon-ui-kit/src/Address/AddressDetails.tsx
```diff
@@ -9,7 +10,8 @@ export type AddressDetailsProps = {
   street: string[];
   postcode?: string;
   city: string;
-  countryId: string;
+  country?: Country;
```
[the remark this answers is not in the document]
nimobeeren: I went with the full `country: Country` prop, since everywhere we use `AddressDetails` we get the address from some `Query` which contains the country object anyway.

packages/falcon-shop-extension/schema.graphql
nimobeeren: I do agree that throwing an error is the best way to indicate that the mutation failed.

As for your example: indeed the return type for a 'new address' mutation should be `Address!`. In normal operation (no errors thrown and mutation is succesful) we always return the address, so the return type is required. 

I think the case of the 'remove address' mutation is similar in that way, so the return type should be `Boolean!` (required). In normal operation, we always return `true`, so the return type is required.

I think the return type should be `Boolean!` instead of `Boolean` since we never explicitly return `null`.

Borales: @nimobeeren we have to be more specific in our comments :) 

Nobody says you need to _ignore_ all returning types. But there are specific cases with Mutations that we don't expect any meaningful value to be returned. 

Like with `setShippingMethod` mutation and its optional `Boolean` return type. There will be no possible cases when this method returns `false`, and even if it does - that does not help us at all, we need a reason why the backend refuses to accept this payload (only possible via `errors` key, and still - the value won't be returned as `false`).

To keep the code cleaner (without having a required `return true` statement) - it could be achieved by making it optional.

In other cases, as you mentioned above - I agree, we need to count on those values.

PR #614, opened 2019-09-30
Query fixes

nimobeeren:

Bug fix.

* ~~Going to a category page (e.g. `/men.html`) crashes.~~
* ~~Adding a customer address (`/account/address-book/add` crashes.~~
* ~~Editing a customer address (`/account/address-book/edit/x`) crashes.~~
* Having an address with street 1 the same as street 2 causes an error about non-unique keys.

Crashes and errors are fixed.

No.

EDIT: I reverted the changes to data extraction from queries, this will be handled by #613. This PR is now only about the change to `AddressDetails`.

### DIFF
--- a/packages/falcon-ui-kit/src/Address/AddressDetails.tsx
+++ b/packages/falcon-ui-kit/src/Address/AddressDetails.tsx
@@ -27,7 +27,7 @@ export const AddressDetails: React.SFC<AddressDetailsProps> = ({
     {company && <Text fontWeight="bold" color="secondaryText">{`${company}`}</Text>}
     <Text fontWeight="bold" color="secondaryText" mb="xs">{`${firstname} ${lastname}`}</Text>
     {street.map((x, i) => (
-      // using array index as key is OK because the order of elements will never change
+      // because `street` can contain non unique values
       // eslint-disable-next-line react/no-array-index-key
       <Text key={i}>{x}</Text>
     ))}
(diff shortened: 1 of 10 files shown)


Borales:
@nimobeeren let's wait until https://github.com/deity-io/falcon/pull/613 is merged (your PR partially contains the same changes)

nimobeeren:
I see, it's essentially the same except for the change to `AddressDetails`

packages/falcon-ui-kit/src/Address/AddressDetails.tsx
```diff
@@ -26,8 +26,10 @@ export const AddressDetails: React.SFC<AddressDetailsProps> = ({
   <AddressDetailsLayout>
     {company && <Text fontWeight="bold" color="secondaryText">{`${company}`}</Text>}
     <Text fontWeight="bold" color="secondaryText" mb="xs">{`${firstname} ${lastname}`}</Text>
-    {street.map(x => (
-      <Text key={x}>{x}</Text>
+    {street.map((x, i) => (
```
tutok: what is the point of this change?

nimobeeren: > Having an address with street 1 the same as street 2 causes an error about non-unique keys.

It's supposed to fix that :point_up:

tutok: thanks! missed that.

But still, I would think if a situation like street 1 is the same as street 2 should be allowed?

packages/falcon-ui-kit/src/Address/AddressDetails.tsx
```diff
@@ -26,8 +26,10 @@ export const AddressDetails: React.SFC<AddressDetailsProps> = ({
...
+      // using array index as key is OK because the order of elements will never change
```
tutok:
```suggestion
      // because `street` can contain non unique values
```

packages/falcon-ui-kit/src/Address/AddressDetails.tsx
nimobeeren: The only thing is that when keys are not unique it prints a big error to console. Nothing breaks but it's a little nicer this way.

PR #627, opened 2019-10-04
Fix for rendering operation errors

### LINKED ISSUE #621: Broken "sign in" - uncaught invariant violation
**Describe the bug**
"Sign In" is broken.

**To Reproduce**
1. Go to http://localhost:3000/
2. Click on "Sign In"
3. Fill any invalid credentials and click "Sign In" under the form
4. See the error in the console

tutok:

fix #621 

**Other information:**


packages/falcon-data/src/Query/OperationError.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ApolloError } from 'apollo-client';
+import { ErrorModel, apolloErrorToErrorModelList } from './apolloErrorToErrorModelList';
+
+export type OperationErrorProps = ApolloError;
+export const OperationError: React.SFC<OperationErrorProps> = props => {
+  const errors = apolloErrorToErrorModelList(props);
+
+  const errorInsights = ({ message, ...rest }: ErrorModel) => {
+    if (process.env.NODE_ENV !== 'production') {
+      return JSON.stringify({ ...rest }, null, 2);
```
Borales: @tutok could we maybe use `Logger` for this? on the client-side - it will use `console.log`, so maybe it would be more useful to get such output in the console.

packages/falcon-shop-data/src/SignIn/SignInMutation.tsx
```diff
@@ -13,7 +14,12 @@ export type SignInResponse = { signIn: boolean };
 export class SignInMutation extends Mutation<SignInResponse, OperationInput<SignInInput>> {
   static defaultProps = {
     mutation: SIGN_IN_MUTATION,
-    awaitRefetchQueries: true,
     refetchQueries: ['Cart', 'CustomerWithAddresses', 'Customer']
   };
 }
+
+export const useSignMutation = (options: MutationHookOptions<SignInResponse, OperationInput<SignInInput>> = {}) =>
+  useMutation(SIGN_IN_MUTATION, {
```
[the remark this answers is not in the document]
tutok: we can but I would not do any changes here because it does not bring any change into types. With current construction, we have less typing definitions because types are inferred from arguments.

packages/falcon-data/src/Query/OperationError.tsx
tutok: here is too late for `Logger`. Introducing `Logger` is not that easy, there is more work than adding one line of code, so we should create a separate task for it.

packages/falcon-server/package.json
```diff
@@ -48,7 +48,7 @@
     "cookies": "~0.7.1",
     "deepmerge": "^2.2.1",
     "eventemitter2": "5.0.1",
-    "graphql": "^14.5.3",
+    "graphql": "14.5.8",
```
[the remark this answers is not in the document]
tutok: your question confusing me, to be honest. Did you forget, we had discussions about this topic some time ago? We agreed was that package dependencies should look like this:
* `dependencies` - all pinned, except `@deity/*` packages
* `devDependencies` - all pinned, except `@deity/*` packages
* `peerDependencies` - semantic

packages/falcon-client/package.json
```diff
@@ -120,7 +120,7 @@
     "@deity/falcon-server": "^2.0.0",
     "@types/fs-extra": "8.0.0",
     "@types/prop-types": "15.7.1",
-    "graphql": "^14.5.3",
+    "graphql": "14.5.8",
```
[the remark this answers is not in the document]
tutok: apollo client version is also pinned
https://github.com/deity-io/falcon/blob/2e5d478804c60240a8c186b818ad722af6d2bb6b/examples/shop-with-blog/client/package.json#L42

packages/falcon-data/src/Query/OperationError.tsx
Borales: @tutok sure, no problem. Closing this one in this case

packages/falcon-server/package.json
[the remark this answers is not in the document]
Borales: I remember that, but we cannot follow this rule blindly all the time.

This is `falcon-server` package, which will be used by the implementation project directly, so my original question was caused by a possible issue with having multiple `graphql` packages installed on the target project.

The original `^14.5.3` was set to match one of the apollo packages' dependencies, so the project would use the same installed package and keep the consistent and uninterrupted flow across the whole application.

I would follow the convention above for dependencies like `deepmerge` or `cookies` (used by FalconServer code internally), but `graphql` (same goes to `graphql-tools`) is one of the critical ones for this project, and the implementation project and its dependencies (like `falcon-server-env`) may use it, so we would have to match the exact version across all of them simultaneously, that's why I'd prefer keeping `graphql` version "open" (with `^`). Besides, if a developer decides to bump `graphql` version - it's his decision, and one change would be applied across all packages.

cc @mateuszwozniak @nimobeeren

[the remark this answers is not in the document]
tutok: so, in short, you want to move `graphql` into peer dependency.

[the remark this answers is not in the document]
Borales: @tutok here's a detailed answer https://github.com/deity-io/falcon/issues/654

packages/falcon-client/package.json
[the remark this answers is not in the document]
tutok: reverted

packages/falcon-server/package.json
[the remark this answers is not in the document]
tutok: reverted

yarn.lock
```diff
@@ -10,6 +10,14 @@
     ts-invariant "^0.4.4"
     tslib "^1.10.0"
 
+"@apollo/react-common@^3.1.3":
```
Borales:
```suggestion
"@apollo/react-common@^3.1.2", "@apollo/react-common@^3.1.3":
```

and remove lines 5-11

yarn.lock
```diff
@@ -21,6 +29,17 @@
     ts-invariant "^0.4.4"
     tslib "^1.10.0"
 
+"@apollo/react-components@^3.1.3":
```
Borales:
```suggestion
"@apollo/react-components@^3.1.2", "@apollo/react-components@^3.1.3":
```

remove lines: 21-30

yarn.lock
```diff
@@ -42,6 +61,16 @@
     ts-invariant "^0.4.4"
     tslib "^1.10.0"
 
+"@apollo/react-hooks@^3.1.3":
```
Borales:
```suggestion
"@apollo/react-hooks@^3.1.2", "@apollo/react-hooks@^3.1.3":
```

remove: 54-62

yarn.lock
```diff
@@ -9541,6 +9570,13 @@ graphql@^14.5.3:
   dependencies:
     iterall "^1.2.2"
 
+graphql@^14.5.8:
```
Borales:
```suggestion
graphql@^14.5.3, graphql@^14.5.8:
```

remove: 9566-9571

examples/shop-with-blog/client/src/pages/Home.js
```diff
@@ -1,36 +1,38 @@
...
+import { PageLayout, ProductList, Loader, OperationError } from '@deity/falcon-ui-kit';
 
 const Home = () => {
-  const { loading, data } = useQuery(GET_CATEGORY_WITH_PRODUCT_LIST, {
-    variables: {
-      categoryId: '25',
-      query: {
-        perPage: 1,
-        page: 20
+  const categoryProducts = () => {
```
mateuszwozniak: To simplify the code I suggest to rename `categoryProducts` to `HomepageProducts` and move it outside (above) `Home` component. Then during final rendering in line 35 instead of doing `{categoryProducts()}` render just functional component `<HomepageProducts />`

packages/falcon-data/src/Mutation/Mutation.tsx
```diff
@@ -1,39 +1,21 @@
...
-          if (!passError && error) {
-            return <Error error={error} />;
-          }
-
-          if (!passLoading && loading) {
-            return <Loader />;
-          }
-          return children(mutation, result);
-        }}
+        {(mutation: MutationFunction<TData, TVariables>, result: MutationResult<TData>) => children(mutation, result)}
```
mateuszwozniak: Why the error/loading states have been removed from here?

packages/falcon-data/src/Query/apolloErrorToErrorModelList.ts
```diff
@@ -0,0 +1,45 @@
...
+import { codes } from '@deity/falcon-errors';
+
+export type ErrorModel = {
+  message: string;
+  code: string;
+  /** path to property (on operation input or operation output) on which error occurs */
+  path?: any;
+};
+
+export const apolloErrorToErrorModelList = (error: ApolloError): ErrorModel[] => {
```
mateuszwozniak: This name (`apolloErrorToErrorModelList`) is very long imo. Maybe we could use something simpler, like `reduceErrors`, `unifyErrors`, `flattenErrors` or something similar?

packages/falcon-shop-data/src/SignIn/SignInMutation.tsx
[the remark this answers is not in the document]
Borales: the warning is gone now 🤷‍♂ , closing this comment

packages/falcon-shop-data/src/Cart/AddToCartMutation.tsx
```diff
@@ -25,3 +26,11 @@ export class AddToCartMutation extends Mutation<AddToCartResponse, OperationInpu
     refetchQueries: ['MiniCart', 'Cart']
   };
 }
+
+export const useAddToCartMutation = (
+  options: MutationHookOptions<AddToCartResponse, OperationInput<AddToCartInput>> = {}
```
Borales:
```suggestion
  options?: MutationHookOptions<AddToCartResponse, OperationInput<AddToCartInput>>
```

can we make the type optional `?:`? (reflecting Apollo's typings, and let Apollo handle the value)

tutok: I would like to say no because suggested change will complicate the `options` merging.

examples/shop-with-blog/client/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@apollo/react-hoc": "^3.1.2",
-    "@apollo/react-hooks": "^3.1.2",
+    "@apollo/react-hooks": "^3.1.3",
```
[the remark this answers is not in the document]
tutok: I already updated, I do not know why yarn still keeps `^3.1.2` (I assume that your question appears because of this)

examples/shop-with-blog/client/src/pages/Home.js
tutok: done

examples/shop-with-blog/client/package.json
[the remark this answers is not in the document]
mateuszwozniak: Line 24 in this file has `^3.1.2`, other files seem to be ok indeed

packages/falcon-data/src/Query/apolloErrorToErrorModelList.ts
mateuszwozniak: Yep, sounds ok to me.

packages/falcon-data/src/Mutation/Mutation.tsx
mateuszwozniak: Ahhh, very good point. That makes the code a bit inconsistent but makes sense anyway from usage point of view.

examples/shop-with-blog/client/package.json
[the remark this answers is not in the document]
mateuszwozniak: Actually searching for `3.1.2` in entire project reveals a few of them ;)

packages/falcon-shop-data/src/Cart/AddToCartMutation.tsx
Borales: Ok, if you think it's better :) 

At least let's make the passing part more consistent, I've seen `, {...options}` and `, options` in other mutations on this PR.

PR #633, opened 2019-10-08
fix: remove unused React imports

DanielRuf:
### Please check if the PR fulfills these requirements ###

- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
This PR removes unused React imports.

**What is the current behavior? (You can also link to an open issue here)**

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)**

**Other information:**

PR #642, opened 2019-10-11
Fix ESLint script

nimobeeren:

npm script fix.

1. `yarn lint` only lints `.js` files (mentioned in #633).
2. pre-commit hook lints entire directory.

1. `yarn lint` lints `.js` files as well as `.ts` and `.tsx`.
2. pre-commit hook only lints staged files.

No.

**Other information**

Though I would like to specify ESLint extensions in a config file, unfortunately setting a command-line option is [the only way](https://eslint.org/docs/user-guide/configuring#specifying-file-extensions-to-lint).

I did not include `.jsx` extension in `lint` script or pre-commit hook, because we do not and probably [should not](https://github.com/facebook/create-react-app/issues/87#issuecomment-234627904) use it.


DanielRuf:
The linting error will be fixed by https://github.com/deity-io/falcon/pull/639/files#diff-5af55d195333df8fe60066776396b6cfR15

mateuszwozniak:
@nimobeeren changes look good, but please fix all errors and warnings caused by these changes. I've just changed #639 to point to your branch as it fixes more than half of the problems. Once it's merged here please take care of the remaining ones.

nimobeeren:
@mateuszwozniak fixed/worked around the last 8 lint errors.

mateuszwozniak:
@nimobeeren thanks, looks good now :)

.lintstagedrc
```diff
@@ -1,4 +1,4 @@
 {
-  "*.{js,jxs,ts,tsx}": ["eslint ./ --fix", "git add"],
+  "*.{js,ts,tsx}": ["eslint --fix", "git add"],
```
tutok: should not we accept also `jsx` files? I am aware that we do not have code in `*.jsx` files but we support them.

```suggestion
  "*.{js,jsx,ts,tsx}": ["eslint --fix", "git add"],
```

(similar change here https://github.com/deity-io/falcon/pull/642/files#diff-dfa8eb367f571d913bbd7ee38a1a5cd0R18  and https://github.com/deity-io/falcon/pull/642/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R17)

mateuszwozniak: I was thinking about that but I don't think we'll be using `jsx` files.

nimobeeren: Did you see the link in the last few lines of the PR?

> I did not include .jsx extension in lint script or pre-commit hook, because we do not and probably [should not](https://github.com/facebook/create-react-app/issues/87#issuecomment-234627904) use it.

I tend to agree with not adding more extensions than needed. In fact, we might want to look into renaming `.tsx` > `.ts`.

DanielRuf: Not sure if this has some side effects because compilers and the toolchain interpret as TSX = TS + JSX. For example the loaders.

nimobeeren: @DanielRuf that might be true indeed. Also `create-react-app` still uses `.tsx` despite that comment from Dan Abramov.

mateuszwozniak: @nimobeeren, after a quick discussion with @tutok we've decided to go anther way - please add `jsx` here as our entire build system covers jsx and we're not going to change that as part of this PR. Once we decide to drop jsx support we'll remove it from here, but until then let's keep things consistent to avoid someone else adding jsx in a few weeks "because our build system handles it".

nimobeeren: Done

Issue #679, opened 2019-10-22
Handle unknown icon


tutok:
**What should be changed/improved?**
if there is no icon component defined in the theme we should notify the developer somehow. We can render some generic placeholder, or throw an error (in development)

right now nothing is rendered, and no error is raised.

EddyVinck:
I would like to do this. Seems like a good entry-level contribution.

PR #682, opened 2019-10-23
improvement: show an error when an icon is not defined in the theme

EddyVinck:

Throw an error when an icon name is used that has not been defined in the theme.

Fixes https://github.com/deity-io/falcon/issues/679

No new behavior, it just adds an error for development.

No.


EddyVinck:
@tutok I tried `throw new Error()` but that caused the `<ErrorBoundary />` to hide the error. I wanted to discuss this with you first so that's why the PR is still a draft.

tutok:
I think `console.error` is fine

packages/falcon-ui/src/components/Icon.tsx
```diff
@@ -24,12 +24,22 @@ export type IconProps = IconRendererProps & {
   fallback?: any;
 };
 const IconInner: React.SFC<IconProps & PropsWithTheme> = props => {
-  if (!props.theme || !props.theme.icons) return null;
+  if (!props.theme || !props.theme.icons) {
```
tutok: please correct me if I am wrong but I think there is no point to check if `!props.theme` exists, as `IconInner` is wrapped via `withTheme`,
similar thing with `!theme.icons`  because of this https://github.com/deity-io/falcon/blob/7a0f4ec654eb0e4f8e584056385ddf1bc36402f4/packages/falcon-ui/src/theme/theme.ts#L118 ?

EddyVinck: I suppose so. The check was already there so I didn't put much thought behind it. Will fix it tomorrow.

EddyVinck: @tutok I just deleted the check. I dug into the code and indeed the check can be deleted because the default theme that is merged in always has an `icons` key.

EddyVinck:
> @EddyVinck @tutok doesn't it deserve to be mentioned in the CHANGELOG?

@Borales I'm glad you brought that up. I didn't find anything about the changelog in the [contribution guidelines](https://github.com/deity-io/falcon/blob/dev/.github/CONTRIBUTING.md). Perhaps some guidelines about what kind of contributions to add to the changelog should be added there as well.

I'll add this change to the changelog.

Borales:
@EddyVinck yeah, you're right, we only have this checkbox in the PR template.. We should probably add it to the guidelines too.

Screenshot 2019-10-24 at 11 28 41

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ Versions marked with a number and date (e.g. Falcon Client v0.1.0 (2018-10-05))
 - fixed theme merging ([#551](https://github.com/deity-io/falcon/pull/551))
 - fixed duplicated conditions when extracting themed css ([#611](https://github.com/deity-io/falcon/pull/611))
 - fixed incompatible `css` prop typings ([#645](https://github.com/deity-io/falcon/pull/645))
+- show an error when an icon is not defined in the theme ([#682](https://github.com/deity-io/falcon/pull/682))
```
Borales:
```suggestion
- added console.error when an icon is not defined in the theme during development ([#682](https://github.com/deity-io/falcon/pull/682))
```

packages/falcon-ui/src/components/Icon.tsx
```diff
@@ -24,12 +24,14 @@ export type IconProps = IconRendererProps & {
   fallback?: any;
 };
 const IconInner: React.SFC<IconProps & PropsWithTheme> = props => {
-  if (!props.theme || !props.theme.icons) return null;
-
   const { icons } = props.theme;
   const { src, fallback, ...rest } = props;
 
   if (!props.src || !icons[src]) {
+    if (process.env.NODE_ENV !== 'production') {
```
nimobeeren: For performance, it might be better to do something like
```
const ENV = process.env.NODE_ENV
...
if (ENV !== 'production) ...
```
where `ENV` is defined outside `IconInner`.

This is because getting the `NODE_ENV` variable is actually more expensive than getting a normal `const`.

nimobeeren: Actually, after some digging it seems like this is not needed when using webpack (https://github.com/OfficeDev/office-ui-fabric-react/issues/8859).
You can disregard this comment.

tutok: I was about replying :) that we actually using `webpack.DefinePlugin`
https://github.com/deity-io/falcon/blob/adc90fff670dc8a87d0e59438f232972ac45993b/packages/falcon-client/build-utils/webpack/config/create.js#L465

your suggestion could be correct but I think only on SSR

tutok: I am worried that you should follow my suggestion, were `!icons` is checked earlier and `null` is returned. The reason is `console.error` will be printed each time when page is loaded.

tutok: @EddyVinck  did you read the above comment?

tutok: here is an issue https://github.com/deity-io/falcon/issues/693 which has an impact on this

EddyVinck: @tutok My bad, I read Nimo's comment about being able to disregard this for now so I thought you implied it will be fixed when we use `DefinePlugin`. I'll add the `const`.

EddyVinck:
@nimobeeren can you review & approve this?

packages/falcon-ui/src/components/Icon.tsx
```diff
@@ -23,13 +25,20 @@ export type IconProps = IconRendererProps & {
...
 const IconInner: React.SFC<IconProps & PropsWithTheme> = props => {
-  if (!props.theme || !props.theme.icons) return null;
+  const { src, fallback, theme, ...rest } = props;
+  const { icons } = theme;
 
-  const { icons } = props.theme;
-  const { src, fallback, ...rest } = props;
+  if (!icons) {
+    return null;
+  }
```
nimobeeren: This check seems redundant to me, and also doesn't use the fallback. I would remove it.

packages/falcon-ui/src/components/Icon.tsx
```diff
@@ -23,13 +25,20 @@ export type IconProps = IconRendererProps & {
...
 
-  if (!props.src || !icons[src]) {
+  if (!src || !icons || !icons[src]) {
+    if (ENV !== 'production') {
+      const errorMessage = `No icon with the name "${src}" was found in your theme.`;
```
Borales: so if no `src` was passed (fits into the first condition of `if` statement) - we're gonna end up by having messages like `No icon with the name "undefined" was found in your theme`? Can we find a better way of generating this msg? (like component path or something)

packages/falcon-ui/src/components/Icon.tsx
tutok: no, it is not redundant, should stay there until https://github.com/deity-io/falcon/issues/693 will be fixed

packages/falcon-ui/src/components/Icon.tsx
tutok: It will never happen, actually check `!src` could be omitted, as `src` is required prop (prop types should be updated)

tutok: `Icon` without `src` param does not make sense, so please do not update the message....

packages/falcon-ui/src/components/Icon.tsx
tutok: more background:
`<LocaleProvider` do the `<Query` (so `<Icon src='loader'` is rendered ) but there is no theme at that moment, as `<ThemeProvider` is defined latte. 
(So my sugestion is: first, we need to fix #693, later we can think about reordering components in tree, if it will be possible.)

image

nimobeeren: Well it is redundant when combined with the `if` statement on line [37](https://github.com/deity-io/falcon/pull/682/files/70865ad18165eeb4e03034249ce373a3479080f4#diff-4cebf9838ac3610923c4ec190028ca5bR37). Either the `if` statement on L33 or the `!icons` on L37 can be removed.

packages/falcon-ui/src/components/Icon.tsx
Borales: so if `if` condition will be updated - then it's fine, the err msg can remain "as is".

EddyVinck: @Borales @tutok so to be clear, you want to remove the `!src` check and add proptypes? I don't see the proptypes that should be updated. There's the `type IconProps` of course, but I don't think that translates to proptypes, right?

tutok: @EddyVinck you are right, `icon` component does not have props types at all, and by saying "prop types should be updated" we think about updating them if they exists or adding them if not, so the final result will be that we will have proper prop types for this component. 
But you can leave it as it is, please decide :).

packages/falcon-ui/src/components/Icon.tsx
tutok: ahh I see right now. So I would leave `if(!icons)` from line `33` but remove `!icons` from `37`

packages/falcon-ui/src/components/Icon.tsx
EddyVinck: Adding some proptypes seems good to me, as the component doesn't make any sense without `src` of course so React should display an error when the prop is omitted.

EddyVinck: image

I just made those changes. I noticed there's an icon `"checkboxCheckmarkIcon"` (see error) that isn't defined in the theme. I don't think this is correct, since that icon is defined in `packages/falcon-ui/docs/themeOverrides.tsx`. Any suggestions how this can be fixed?

tutok: @EddyVinck I am looking at it, will provide a solution soon 😃

tutok: ok, I looked at this closer and I think `IconInner` should look like this:
```
const IconInner: React.SFC<IconProps & PropsWithTheme> = props => {
  const { src, fallback, theme, ...rest } = props;
  const { icons } = theme;

  // TODO: temporary fix for https://github.com/deity-io/falcon/issues/693, remove when resolved
  if (!icons) {
    return null;
  }

  if (icons[src]) {
    const { icon, ...otherProps } = icons[src];

    return <IconRenderer as={icon} {...(otherProps as any)} {...rest} />;
  }

  if (fallback) {
    return fallback;
  }

  if (ENV !== 'production') {
    console.error(`There is no icon "${src}" in your theme ("theme.icons"), nor icon fallback is defined.`);
  }

  return null;
};
```

please look at it.

EddyVinck: Looks good to me @tutok ! :+1: 
I'll push that to this PR

EddyVinck: @tutok I tested it and it fixes the issue with the fallback icon. The commit has been pushed, so I think this PR is able to be merged now.

Borales: @tutok @EddyVinck resolving this thread. If needed - we can open a new one

Borales:
@EddyVinck please merge the latest changes from `dev` branch, you have conflicts

packages/falcon-ui/src/components/Icon.tsx
```diff
@@ -23,19 +26,37 @@ export type IconProps = IconRendererProps & {
...
+    );
+  }
+
+  return null;
 };
 
 export const Icon = withTheme(IconInner) as React.SFC<IconProps>;
+
+Icon.propTypes = {
+  src: PropTypes.string.isRequired
```
Borales: @EddyVinck @tutok just to make it clear - `src` is just a **name** of the icon in our `theme.icons` dictionary, it not actually refers to a normal "source" of the file

tutok: yes, `src` refer to key in `[your-theme].icons` object

EddyVinck: @Borales when you use the Icon component, the `src` prop refers to the icon key of the icon in the `theme.icons` map.

Issue #601, opened 2019-09-23
Switch to Terser


DanielRuf:
**What should be changed/improved?**
Currently the old UglifyJS plugin is use but it is slow and buggy.

https://github.com/deity-io/falcon/blob/b1ff5b47b89f986c62b312426d3278a24e97b2e1/packages/falcon-client/build-utils/webpack/config/create.js#L4

**Your suggested solution**
Switch to https://github.com/webpack-contrib/terser-webpack-plugin

PR #683, opened 2019-10-23
improvement: switch from uglify to terser webpack plugin

EddyVinck:

Changes UglifyJsPlugin to TerserWebpackPlugin https://github.com/deity-io/falcon/issues/601

The bahavior is the same

The new behavior is the same, except Terser is maintained

I tested it by running `yarn start` in the root directory, but I would like some help to verify that nothing is broken.

**Other information:**
Terser is a fork of Uglify, like Daniel said. I checked and the API should be compatible. All the options used in Falcon are supported in the same configuration structure by Terser.


packages/falcon-client/package.json
```diff
@@ -130,7 +130,8 @@
     "react-dom": "16.10.2",
     "react-router-dom": "5.0.1",
     "react-test-renderer": "16.10.2",
-    "supertest": "4.0.2"
+    "supertest": "4.0.2",
+    "terser-webpack-plugin": "^2.2.1"
```
tutok: please remove `uglifyjs-webpack-plugin`

packages/falcon-client/build-utils/webpack/config/create.js
```diff
@@ -487,8 +487,8 @@ module.exports = (target = 'web', options) => {
         ...config.optimization,
         minimize: true,
         minimizer: [
-          new UglifyJsPlugin({
-            uglifyOptions: {
+          new TerserPlugin({
+            terserOptions: {
```
tutok: looks like option `terserOptions.parallel` is set by default into `true`, so maybe we can remove it (with its comment too)

---
I also did a review of other settings and it seems that they need to stay as they were, `create-react-app` uses similar settings.

tutok:
`yarn build:prod` produces `5KB` larger bundle, so no improvements 😞

EddyVinck:
> `yarn build:prod` produces `5KB` larger bundle, so no improvements

larger than when using uglify? t's for the better I suppose but that's kind of sad :(

packages/falcon-client/package.json
tutok: I think you should also run `yarn install` to update `yarn.lock`.
If you will not do this right now, someone could fetch your changes latter (no matter if from this branch or latter from `dev`) and after `yarn install`, he will have local changes.

tutok:
@EddyVinck  can you please resolve conflicts? after that I think we can merge

EddyVinck:
@tutok this should be good to go now

PR #686, opened 2019-10-24
Re-exporting Falcon-Server components/utils

Borales:

Improvement (to be able to re-use internal utils and components in Falcon-Server based apps).


packages/falcon-server/src/index.ts
```diff
@@ -30,15 +30,24 @@ import get from 'lodash/get';
...
+import { cacheInvalidatorMiddleware } from './middlewares';
+import { GraphQLCacheDirective, GraphQLCacheInvalidatorDirective } from './schemaDirectives';
 import { Config, BackendConfig } from './types';
 
+export * from './containers';
+export * from './graphqlUtils';
+export * from './middlewares';
+export * from './resolvers';
+export * from './schemaDirectives';
+export * from './utils';
```
tutok: I am not sure if we should export everything from `falcon-server` (which as I knew) should be considered as a tool not like a library. I think things used internally via `falcon-server` should not be exposed, If we need some its internal I think the right way is to move it into `falcon-server-env` (which is as I know, is the set of prerequisites for `falcon-server`)

Borales:
@tutok @nimobeeren `falcon-server/graphqlUtils` module has been moved to `falcon-server-env` for reusability

packages/falcon-server/src/index.ts
Borales: done

tutok: so should not we remove this line?

Borales: removed

PR #687, opened 2019-10-24
Improved gitignore

tutok:

improvement

development configuration should be in repository, to make sure that all of us have the same config

no

**Other information:**
no

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,10 +14,14 @@ package-lock.json
 *.swp
 *.swo
 
-
 packages/**/dist/
+
 examples/**/build/
-examples/**/public/i18n/
+examples/**/config/*
+!examples/**/config/custom-environment-variables.*
+!examples/**/config/default.*
+!examples/**/config/development.*
+!examples/**/config/test.*
 
 #sandbox:
 examples/**/static/dist/
@@ -25,6 +29,4 @@ examples/**/static/manifest.json
 examples/**/webpack-assets.json
 examples/**/webpack-stats.json
 
-examples/**/config/*
-!examples/**/config/custom-environment-variables.*
-!examples/**/config/default.*
+
--- a/examples/shop-with-blog/server/config/development.json
+++ b/examples/shop-with-blog/server/config/development.json
@@ -1,0 +1,9 @@
+{
+  "debug": true,
+  "logLevel": "trace",
+  "cache": {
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -14,17 +14,12 @@ package-lock.json
 *.swp
 *.swo
 
-
 packages/**/dist/
-examples/**/build/
-examples/**/public/i18n/
```
Borales: @tutok `i18n` is not being copied to the `public` folder anymore?

tutok: it is into `/build/public` instead of `/public`

.gitignore
```diff
@@ -14,17 +14,12 @@ package-lock.json
...
-examples/**/static/dist/
-examples/**/static/manifest.json
-examples/**/webpack-assets.json
-examples/**/webpack-stats.json
 
+examples/**/build/
 examples/**/config/*
 !examples/**/config/custom-environment-variables.*
 !examples/**/config/default.*
+!examples/**/config/development.*
```
nimobeeren:
```suggestion
!examples/**/config/development.*
!examples/**/config/production.*
```

This file is tracked so it probably shouldn't be ignored?

Borales: `production` files may contain some very sensitive information, so I wouldn't keep it in the GIT simply to avoid such cases. Besides, this is a template app, if a developer/agency really needs this - they could change it any time.

Borales: current we do have this file in the git as a template file, but any changes to it are being ignored

nimobeeren: As far as I know, when a file is tracked, git detects changes even though it is ignored. Indeed, when I change `shop-with-blog/client/config/production.json`, it shows up in `git status`.

I am not aware of a way to create such a template file. There is `update-index --assume-unchanged` but I believe that only works locally.

Unless you know a way to ignore changes while keeping it tracked, I suggest we remove the `production.json` file and keep it ignored.

Borales: @nimobeeren heh, you're right, something went wrong :) but still, I wouldn't encourage people to keep it in git (at least on the `client/.gitignore` level)

nimobeeren: @Borales If you agree, could you remove `production.json` from git?

tutok: @nimobeeren  you are right, I think I need to remove `production.json` from repo

Issue #688, opened 2019-10-24
falcon-scripts command keeps printing error msgs after the command is done


Borales:
**Describe the bug**
When building `falcon-server` package - I've noticed that `falcon-scrips build` keeps printing errors after the initial command is done when having rejections for more than 1 Promise.

**To Reproduce**
1. Open any file in `falcon-server` package
2. Make a typo in the import statement, like `import { generateTagNames } from '../graphqlUtils2';` in `packages/falcon-server/src/middlewares/cacheInvalidatorMiddleware.ts`
3. See error:

```
yarn run v1.19.1
$ npm run clean && npm run build

> @deity/falcon-server@2.0.0 clean /Volumes/dev/deity/falcon/packages/falcon-server
> falcon-scripts clean


> @deity/falcon-server@2.0.0 build /Volumes/dev/deity/falcon/packages/falcon-server
> falcon-scripts build

building dts...
building esm...
building cjs...
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Successfully compiled 17 files with Babel.
'path' is imported by src/index.ts, but could not be resolved – treating it as an external dependency
'fs' is imported by src/index.ts, but could not be resolved – treating it as an external dependency
'path' is imported by src/containers/BaseContainer.ts, but could not be resolved – treating it as an external dependency
'path' is imported by src/containers/ExtensionContainer.ts, but could not be resolved – treating it as an external dependency
'fs' is imported by src/containers/ExtensionContainer.ts, but could not be resolved – treating it as an external dependency
Error: Could not resolve '../graphqlUtils2' from src/middlewares/cacheInvalidatorMiddleware.ts
    at error (/Volumes/dev/deity/falcon/node_modules/rollup/dist/rollup.js:9435:30)
    at ModuleLoader.handleMissingImports (/Volumes/dev/deity/falcon/node_modules/rollup/dist/rollup.js:16504:17)
    at ModuleLoader.<anonymous> (/Volumes/dev/deity/falcon/node_modules/rollup/dist/rollup.js:16555:26)
    at Generator.next (<anonymous>)
    at fulfilled (/Volumes/dev/deity/falcon/node_modules/rollup/dist/rollup.js:40:28) {
  code: 'UNRESOLVED_IMPORT'
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @deity/falcon-server@2.0.0 build: `falcon-scripts build`
npm ERR! Exit status 1
npm ERR!
[...]

PR #691, opened 2019-10-25
fix for falcon-scripts command error messages printing after its done

tutok:

fixed https://github.com/deity-io/falcon/issues/688

**Other information:**


Borales:
[discussion]
@tutok previously, it was a Rollup's exception, now it seems like it comes from `cross-spawn` - that's because we handle such exceptions properly now, correct? 

Another thing - since this is a cross-spawn's exception - I don't think it's very useful to render a complete `error` object (showing a trace-stack from the cross-spawn package), maybe we could show [here](https://github.com/deity-io/falcon/pull/691/files#diff-b5bee407fcbc50aabb9e4211659713eeR77) `error.message` only? (it contains file names as well)

packages/falcon-dev-tools/falcon-scripts/bin/falcon-scripts.js
```diff
@@ -18,20 +21,28 @@ const test = require('../src/test');
...
-            fs.existsSync(paths.pkgBinSrc) && require('../src/build-cjs').bin({ packagePath })
+            dts.build({ packagePath }),
+            esm.build({ packagePath }),
+            cjs.main({ packagePath }),
+            fs.existsSync(paths.pkgBinSrc) && cjs.bin({ packagePath })
           ].filter(x => x)
         );
 
+        if (results.some(x => !x.isSuccess)) {
+          throw results.find(x => !x.isSuccess).error;
```
Borales: @tutok does this mean it will re-throw only the first error? Having more than 1 error - it'll require another re-run to be able to see it, right? Can we render all available errors at once?

tutok: Personally I think this is ok because in most cases `build` "scripts" (esm, cjs, dts) could fail at the same time for the same reason? so what is the reason to print the same error more than one time?

Moreover, if we want to print more erros we should improve printing output in generall to rewrite it and add some kind of prefix

tutok:
@Borales because the order of reported exceptions was changed. `cross-spawn-promise` is wrapping an error, so if we do not want it, I need to remove this dependency.

packages/falcon-dev-tools/falcon-scripts/bin/falcon-scripts.js
mateuszwozniak: That's true, 99% of cases will fail for all processors on the same problem, so it's fine to rethrow only the first one.

Borales: Ok, got it.

What about printing `error.message` only? (I guess 99% of the cases - the stack will come from `cross-spawn`, not very helpful). I checked locally - looks more neatly.

tutok: `esm` build is created by `cross-spaw-promise` and errors from this script will be wrapped by this package (the stack will `contain cross-spaw-promise`), but in the same time `cjs` is the raw rollup execution, so no `cross-spaw-promise` in stack

In my opinion, your idea is bad, because the code responsible for printing errors will need to know how the script was invoked? sounds weird. I think the better will be when we moved error handing inside script, but In my opinion, the proper solution is to throw away `cross-spaw-promise`. 

I will do the last, as I think you do not like `cross-spaw-promise`

Issue #647, opened 2019-10-15
Add Falcon to Wappalyzer


EddyVinck:
**What should be changed/improved?**
Wappalyzer is a chrome extension with approximately 1 million installs. It allows users to see what technologies a website is using by checking out the website's HTML and by analyzing the requests it sends.

https://www.wappalyzer.com/docs/specification

https://chrome.google.com/webstore/detail/wappalyzer/gppongmhjkpfnbhagpmjfkannfbllamg

**Your suggested solution**
If we add Falcon to their Chrome extension people can see if a website is built with Falcon. They might even check it out if they've never seen it before.

PR #692, opened 2019-10-25
improvement: add a generator meta tag to check if someone is using Falcon

EddyVinck:

fixes https://github.com/deity-io/falcon/issues/647

Currently there is no easily identifiable aspect of Falcon apps. 

This PR adds a generator meta tag we can use to add Falcon to Wappalyzer and other tools.

No.

**Other information:**
Of course users might override this meta tag with something else, so it's not a "bulletproof" solution.


packages/falcon-client/src/components/HtmlHead.js
```diff
@@ -3,7 +3,11 @@ import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet-async';
 
 export default function HtmlHead({ htmlLang }) {
-  return <Helmet htmlAttributes={{ lang: htmlLang }} />;
+  return (
+    <Helmet htmlAttributes={{ lang: htmlLang }}>
+      <meta name="generator" content="DEITY Falcon" />
```
Borales: @EddyVinck maye we could put it here directly? (without using helmet)
https://github.com/deity-io/falcon/blob/dev/packages/falcon-client/src/components/Html.js#L33

mateuszwozniak: Yes, I think it's a better option, but wouldn't it be even better to add a special header instead of meta tag?

EddyVinck:
> The meta tag is added correctly to the DOM. Now I suppose we have to let Wappalyzer know about Falcon before it can be recognized?

Correct. We still need to inform 3rd party application like Wappalyzer what to look for.

PR #695, opened 2019-10-28
Fixing `@types/prop-types` package across the project

Borales:
The main issue is that PropTypes validation throws false-negative warnings on any usage of `.shape()` (deep validation) method. Using `.exact()` seems to be a temporary solution until the real issue is solved.

Screenshot 2019-10-28 at 22 08 55

Reference issue:
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34692


packages/falcon-data/src/Mutation/Mutation.tsx
```diff
@@ -12,9 +12,16 @@ export class Mutation<TData = any, TVariables = OperationVariables> extends Reac
...
+    // node_modules/@apollo/react-components/lib/Mutation.d.ts:9
+    // Compiler throws an error with the following message:
+    // --------------------------------------------------------------------------------
+    // Type 'TData | ((vars: TVariables) => TData)' is not assignable to type 'object'.
+    // Type 'TData' is not assignable to type 'object'
+    // --------------------------------------------------------------------------------
+    const { children, optimisticResponse, ...restProps } = this.props;
     return (
-      <ApolloMutation {...restProps}>
+      <ApolloMutation optimisticResponse={optimisticResponse as any} {...restProps}>
```
Borales: I'd be more than happy to replace this hack with a proper fix.

nimobeeren:
Wouldn't `.exact()` be the correct solution regardless? The TS `type` is enforcing no unknown properties in the `image` object, which is what `.exact()` wants. Using `.shape()` would be a weaker constraint than necessary.

Borales:
@nimobeeren `prop-types` checks the shape of the data during the runtime, while TS does it during the compilation. So it could be a problem when extending the object with custom data, like for `ProductCard` - I'm getting this logs in the console:

```
Warning: Failed prop type: Invalid prop `price` key `minTier` supplied to `ProductCard`.
Bad object: {
  "regular": 29,
  "special": null,
  "minTier": null,
  "__typename": "ProductPrice"
}
Valid keys: [
  "regular",
  "special"
]
```

so I have to find a way to fix this

nimobeeren:
@Borales hmm yeah it can be annoying when you pass in an object with keys that you don't want to use (like `__typename`). But, when possible, the prop types should include keys like `minTier` if it knows that it wants to use that.

I think in most cases we know the exact shape of the prop so then that should be reflected in both TS and `prop-types`.

Borales:
@nimobeeren I submitted an [issue](https://github.com/facebook/prop-types/issues/296) to `prop-types` repo, and revered the code to original methods (with `@ts-ignore` comment). Once the issue is resolved - I'll remove them.

https://codesandbox.io/s/prop-types-ts-validation-failure-f1h4n with reproduction.

It's not just `@types/prop-types` issue, so I don't see any other solution right now.

packages/falcon-ui-kit/src/Blog/BlogPostExcerpt.tsx
```diff
@@ -38,6 +38,7 @@ BlogPostExcerpt.propTypes = {
   slug: PropTypes.string.isRequired,
   date: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)]).isRequired,
   excerpt: PropTypes.string.isRequired,
+  // @ts-ignore https://github.com/facebook/prop-types/issues/296
```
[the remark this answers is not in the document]
tutok: my bad, I understand it now

packages/falcon-ui-kit/src/Locale/LocalePicker.tsx
```diff
@@ -29,8 +29,9 @@ export const LocalePickerInnerDOM: React.SFC<LocalePickerProps> = ({ items, valu
   </Dropdown>
 );
 LocalePickerInnerDOM.propTypes = {
+  // @ts-ignore https://github.com/facebook/prop-types/issues/296
   items: PropTypes.arrayOf(
-    PropTypes.exact({
+    PropTypes.shape({
```
tutok: so this should be reverted i think 
```suggestion
    PropTypes.exact({
```

PR #696, opened 2019-10-28
Fix example client dependencies and add "build:prod" to the server

mateuszwozniak:

1. Moved most of the dependencies from "dependencies" to "devDependencies" in `client/package.json` as `yarn install --production` should install only required packages to run in production mode, nothing more
2. Added `build:prod` step in `server/package.json` so it can be used in Dockerfiles if some additional work needs to be done before running FS in production mode


Borales:
@mateuszwozniak I think we should also update https://falcon.deity.io/docs/deployment/intro page with an explanation for this step

mateuszwozniak:
I think we'll update it during this one: #664 - it has just development docker files atm, but we'll have to add production files plus information how everything should be used together

Issue #693, opened 2019-10-25
locales from BackendConfig were not merged with locales from ClientConfig


tutok:
**Describe the bug**
when page is rendered we use `clientConfig`, latter we fetch `backendConfig` via `LocaleProvider`, and we simply overwrite locales and currency. We overwrite even default language. So `clientConfig` is totally ignored. 

Moreover `htmlLang` attribute is not updated when language is changed because its value is taken from `clientConfig` - `backendConfig` does not have any impact on it. 

**Expected behavior**
we should merge both configs and use only languages supported by the client app and backend.

**Additional context**
I think we should consider fetching `backendConfig` on server (when SSR)

PR #698, opened 2019-10-30
Fix for localization handling

tutok:

fix: https://github.com/deity-io/falcon/issues/693

**Other information:**


packages/falcon-front-kit/src/Locale/LocaleContext.tsx
```diff
@@ -24,34 +25,37 @@ export const LocaleProvider: React.SFC<LocaleProviderProps> = ({
...
+    {({ data: { clientConfig } }) => (
+      <BackendConfigQuery>
+        {({ data: { backendConfig } }) => {
+          const { activeLocale, shop } = backendConfig;
 
-      const locale = activeLocale || 'en';
-      const localeFallback = 'en';
-      const currency = props.currency || shop.activeCurrency || 'EUR';
+          const localeFallback = clientConfig.i18n.lng;
+          const currency = props.currency || shop.activeCurrency || 'EUR';
```
Borales: @tutok just a minor **question** - shouldn't we keep the currency outside of this context? Maybe a separate `ShopContext`? not every Falcon-based app needs currency data as it may not use it at all (including `priceFormat`). 

cc @mateuszwozniak

tutok: Actually I thought about this some time ago. and I can agree that it could make sense to extract something like `<CurrencyProvider />`.

mateuszwozniak: Good point, fully agree.

Borales: @tutok that would work, thanks!

Issue #561, opened 2019-08-27
deduplicate dependencies in yarn.lock


tutok:
**What should be changed/improved?**
`yarn.lock` contains tons of duplicated packages, we could optimize it.

**Your suggested solution**
probably we can automate it via [yarn-deduplicate](https://www.npmjs.com/package/yarn-deduplicate#deduplication-strategies) (run always on `yarn prepare`)

tutok:
when worked on #558 I tried to incorporate yarn-deduplicate but after a quick try, it looks like it does not work deterministically. Please try it one more time.

PR #700, opened 2019-10-30
Add yarn deduplicate script

tutok:

improvement

`yarn install` will install new dependencies and potentially create some duplications

`yarn install` will install new dependencies, and `npx yarn-deduplicate yarn.lock` on `postinstall` script will be executed in order to remove potential duplications added by the former command

no

**Other information:**
no


package.json
```diff
@@ -6,13 +6,15 @@
...
     "new-package": "node scripts/new-package.js",
+    "deduplicate": "npx yarn-deduplicate yarn.lock",
     "cleanup": "rimraf packages/**/{dist,coverage} packages/create-falcon-app/examples ./**/node_modules",
     "cleanup:all": "rimraf yarn.lock && yarn cleanup",
     "test": "lerna run test:coverage",
     "coverage": "codecov -p packages",
     "watch": "lerna run watch --parallel",
     "start": "cross-env CI=true lerna run start:dbg --scope=@deity/example-shop-with-blog* --parallel",
     "prepublish": "rimraf node_modules/@emotion/core/types && lerna run --scope=@deity/falcon-ui --stream prepare",
+    "postinstall": "yarn deduplicate",
```
Borales: I wasn't sure if it would be safe to add such command while working on #561, so I added https://github.com/deity-io/falcon/wiki/yarn.lock#resolving-duplicates wiki page to help others. 

If we go this way - we may want to drop such a check on CI (https://github.com/deity-io/falcon/blob/dev/.github/workflows/main.yml#L26-L28) because it won't make any difference (especially considering the upcoming limits for CI minutes), `postinstall` will do the job.

tutok: >To resolve `yarn.lock` conflicts and/or remove unwanted package updates - you need to:
> * run `yarn cleanup` within Falcon root folder
> * replace your local `yarn.lock` with [this one](https://raw.githubusercontent.com/deity-io/falcon/dev/yarn.lock) (from `dev` branch)
> * run `yarn` to make sure Yarn will check the latest changes from other packages once again and update `yarn.lock` if needed

so is it not the same as accepting incoming changes when any conflicts, and `yarn install`?

at `postinstall` - do you want to use `yarn-dedupe` in order to list duplicates only, and remove them manually?

Borales: @tutok not that. Actually, on the second that - maybe more effectively would be doing something like this in `.lintstagedrc`:

```diff
{
  "*.{js,jsx,ts,tsx}": ["eslint --fix", "git add"],
  "*.{json,css,scss,graphql,yml}": ["prettier --write", "git add"],
+  "yarn.lock": ["npx yarn-deduplicate", "git add"]
}
```

This way we don't trigger `npx yarn-deduplicate` on installs without actual updates and deduplicate the deps before committing the changes (this way - we can "ensure" those deduplicated changes will be committed). 

I'm just not a big fan of doing such thing on `postinstall`, because this event will be triggered on the CI as well, and potentially deduplicating the dependencies before running the tests - it could introduce an inconsistency.

What do you think @tutok @mateuszwozniak ?

tutok: > I'm just not a big fan of doing such thing on postinstall, because this event will be triggered on the CI as well, and potentially deduplicating the dependencies before running the tests - it could introduce an inconsistency.

make sense

nimobeeren: @Borales do we still need `npx` if `yarn-deduplicate` is installed as a dependency?

Borales: @nimobeeren it's updated  👍

package.json
```diff
@@ -6,13 +6,15 @@
   "repository": "git@github.com:deity-io/falcon.git",
   "scripts": {
     "new-package": "node scripts/new-package.js",
+    "deduplicate": "yarn yarn-deduplicate yarn.lock",
```
Borales:
```suggestion
    "deduplicate": "yarn-deduplicate yarn.lock",
```

it can be omitted

PR #720, opened 2019-11-12
Improving data-sharing across Containers

Borales:

Improvement

When implementing endpoints for cache invalidation - there was no `cache` component exposed in the `EndpointManager` instance.

**Does this PR introduce a breaking change?**
No

### DIFF
--- a/packages/falcon-server/src/containers/EndpointContainer.ts
+++ b/packages/falcon-server/src/containers/EndpointContainer.ts
@@ -1,10 +1,15 @@
 /* eslint-disable no-restricted-syntax, no-await-in-loop, no-underscore-dangle */
-import { EndpointEntry, EndpointConstructor } from '@deity/falcon-server-env';
+import { Cache, EndpointEntry, EndpointConstructor } from '@deity/falcon-server-env';
+import { EventEmitter2 } from 'eventemitter2';
 import { EndpointEntryMap } from '../types';
 import { BaseContainer } from './BaseContainer';
 
 export class EndpointContainer extends BaseContainer {
   public entries: Array<EndpointEntry> = [];
+
+  constructor(eventEmitter: EventEmitter2, protected cache: Cache) {
+    super(eventEmitter);
+  }
 
   /**
    * Instantiates endpoints based on the passed configuration and registers event handlers for them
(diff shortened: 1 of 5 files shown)


packages/falcon-server/src/containers/EndpointContainer.ts
```diff
@@ -21,6 +26,7 @@ export class EndpointContainer extends BaseContainer {
           return;
         }
         const endpointManager = new EndpointManagerClass({
+          cache: this.cache,
```
nimobeeren: Perhaps I'm missing something but I don't think `this.cache` is defined here. The `cache` property was added to the `EndpointManager` class but this class does not inherit from it, nor do we set its value in the constructor.

Borales: @nimobeeren `cache` is being passed on line 10 of this file. `EndpointManager` constructor argument has been updated as well:

Screenshot 2019-11-14 at 09 44 50

nimobeeren: @Borales the constructor of `EndpointContainer` does take a `cache` parameter, but I don't see that we use it anywhere. Yet we try to access `this.cache` without ever setting it.

Borales: @nimobeeren I guess you're asking about the constructor assignment, aren't you? 

```typescript
constructor(eventEmitter: EventEmitter2, protected cache: Cache) {

}
```

Having `protected` keyword in the constructor argument automatically does the assignment for us.

Borales: Here's how it looks after being compiled - [Playground Link](http://www.typescriptlang.org/play/index.html?target=1#code/MYGwhgzhAEBiD29oG8CwAoaXrHgOwgBcAnAV2EPmIAoAHY+QgUwqYBNo8wBbJgLmhFiASzwBzAJQoM2WdBnYAvhkVA)

nimobeeren: @Borales ah cool! I was looking for something like that but couldn't find it in the TS docs. That clears it up!

PR #722, opened 2019-11-19
Enabling falcon-front-kit to export SearchContext + dependency update

Borales:
- Re-exporting `SearchContext` to enable calling `useContext(SearchContext)`
- Updating "react-lazyload" dependency to fix the image loading

### DIFF
--- a/packages/falcon-front-kit/src/Search/useSearchContext.ts
+++ b/packages/falcon-front-kit/src/Search/useSearchContext.ts
@@ -1,4 +1,4 @@
 import { useContext } from 'react';
 import { SearchContext } from './SearchContext';
 
-export const useSearchContext = useContext(SearchContext);
+export const useSearchContext = () => useContext(SearchContext);
--- a/packages/falcon-front-kit/src/Search/useSearchContext.ts
+++ b/packages/falcon-front-kit/src/Search/useSearchContext.ts
@@ -1,0 +1,4 @@
+import { useContext } from 'react';
+import { SearchContext } from './SearchContext';
+
+export const useSearchContext = useContext(SearchContext);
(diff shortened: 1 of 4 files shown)


packages/falcon-front-kit/src/Search/useSearchContext.ts
```diff
@@ -0,0 +1,4 @@
+import { useContext } from 'react';
+import { SearchContext } from './SearchContext';
```
tutok: We should not export `SearchContext` as it is not a part of the public API. It is an internal dependency of Providers and Consumers (which could be considered as public).  

```suggestion
```

Borales: @tutok you mean not export it at all (for hooks) or just `SearchContext`?

Borales: `SearchContext` is not being exported

tutok: I am lost, did I add comment into `import`? Please forgive me! Right now it is looks ok!

I mean not making public `SearchContext` as it is an internal dependency of consumer and provider. `useContext` creates consumer hook so it is reasonable to export.

Borales: 👍 all good!

packages/falcon-ui-kit/package.json
```diff
@@ -29,7 +29,7 @@
     "core-js": "3.2.1",
     "history": "4.9.0",
     "react-credit-card-input": "1.1.5",
-    "react-lazyload": "2.6.2",
+    "react-lazyload": "^2.6.5",
```
nimobeeren: Should we use pinned version here since we are doing that for all other (non-falcon) deps?

Borales: @nimobeeren that's under discussion right now, but to avoid such problems in the future (there were several patches for this package with the bugfix, but it was impossible to use it because of the hardcoded version in our package) - we might think of using `^` for dependencies to unblock end-developers from waiting for fixes from our end.

PR #743, opened 2019-12-03
ShopExtension - adding Category fields

### LINKED ISSUE #653: Shop-extension Category type is missing important fields
When requesting a specific Category on GQL - `urlPath` is not present:

```graphql
query Category {
  category(id:"19") {
    id
    name
  }
}
```

It would be nice to have:
- `urlPath`
- `image`
- `attributes`

Borales:
Adding basic fields to Category type in Shop Extension schema (and resolvers for magento2-api)

Improvement

No

**Other information:**
Fixes #653


packages/falcon-front-kit/src/SEO/SEO.tsx
```diff
@@ -0,0 +1,30 @@
...
+import PropTypes from 'prop-types';
+import { Helmet } from 'react-helmet-async';
+import { SEO as SEOType } from '@deity/falcon-shop-extension';
+
+export type SeoProps = {
+  meta: Partial<SEOType>;
+  title: string;
+};
+
+export const SEO: React.FC<SeoProps> = ({ meta: { title: metaTitle, description, keywords }, title }) => (
```
nimobeeren: Why do we need both `title` and `meta.title`?

Borales:
@nimobeeren I pushed the fix (weird, it wasn't related to my changes). Btw, the query should have another nested level:

```graphql
productList {
  items {
    id
    categories {
      name
    }
  }
}
```

packages/falcon-front-kit/src/SEO/SEO.tsx
Borales: @nimobeeren `title` is something that a developer needs to provide manually (and it's a required prop). `meta` may contain a custom `meta.title` (set for SEO-purposes), but if `meta` doesn't provide one - `title` will be used for this purpose, so we'll be having at least 1 required title.

packages/falcon-magento2-api/src/index.ts
```diff
@@ -320,8 +326,9 @@ module.exports = class Magento2Api extends Magento2ApiBase {
    * @returns {Promise<Product[]>}  response with list of products
    */
   async productList(obj, { input }) {
-    const { withAttributeFilters = [] } = input;
-    const searchCriteria = this.createSearchCriteria(input);
+    // It's not being used anymore.
+    const withAttributeFilters = [];
```
mateuszwozniak: So just remove this variable and put `withAttributeFilters: []` in the actual call (if Magento requires it).

Borales: removed

packages/falcon-shop-extension/src/types.ts
```diff
@@ -334,7 +335,7 @@ export type ConfigurableProductOptionValue = {
   valueIndex: string;
 };
 
-export type ProductSeo = {
+export type SEO = {
```
tutok: should not we add also fields like `image` or `urlPath`? these are also necessary from SEO perspective

examples/shop-with-blog/client/src/pages/shop/Category/Category.js
```diff
@@ -38,9 +37,7 @@ const CategoryPage = ({ match: { params } }) => (
 
           return (
             <CategoryLayout variant={!filtersData.length && 'noFilters'}>
-              <Helmet>
-                <title>{name}</title>
-              </Helmet>
+              <SEO meta={category.seo} title={name} />
```
tutok: > should not we add also fields like image or urlPath? these are also necessary from SEO perspective

(referring to this comment)

`category` (I think `Product` or `BlogPost` also) contains more SEO information like mentioned above `image`. So maybe we could introduce interface (`ISeoEntity`) which could be later implemented via `Category`, `Product`, `BlogPost`... then above line could look like:
```suggestion
              <SEO {...category} />
```

I think Interface is better than duplicate all fields into `entity.seo` node

packages/falcon-shop-extension/src/types.ts
Borales: @tutok this task was mainly focused on adding missing fields to Category, and in order to re-use some existing types - I renamed `ProductSeo` to `SEO` (it wouldn't make sense to use "Product" prefix for Category).

There's even much more stuff related to SEO for products, like price, stock, reviews, brand etc - but I think this deserves to have a separate task to fully cover SEO needs

examples/shop-with-blog/client/src/pages/shop/Category/Category.js
Borales: @tutok the proposed idea is great, but I'm afraid merging fields into the main type could lead to unwanted behavior like having a need to use a custom title for `<title>` tag (for example "Buy these cool BrandName X100 headphones"), while keeping `<h1>` more simple for visitors (for example "BrandName X100 headphones") - this way `title` is not a duplicate and serves different purposes. Same goes to `description` - `description` of the Category may contain a fully-featured HTML content, while `seo.description` is limited (by length) and must be a plaintext.

Besides that, keeping SEO-related stuff under `seo` key lets other developers using this data from another source (like external SEO-tool) while keeping the main Product/Category data to be fetched from the shop-backend.

PR #751, opened 2019-12-12
Optional "move forward" on Checkout

tutok:

improvement

checkout `step` is changed automatically when its `values` are updated

it is possible to configure default behavior, so checkout `step` can be updated via `stepForward()` method

**Other information:**

### DIFF
--- a/packages/falcon-front-kit/src/Checkout/CheckoutProvider.tsx
+++ b/packages/falcon-front-kit/src/Checkout/CheckoutProvider.tsx
@@ -1,7 +1,7 @@
 import React, { useState, useCallback } from 'react';
 import { PlaceOrderResult } from '@deity/falcon-shop-extension';
 import { CheckoutValues, SetCheckoutValues } from './CheckoutValues';
-import { CheckoutStep, CheckoutStepType, CheckoutFlow, getNextStepForValues, getNextStepFactory } from './CheckoutStep';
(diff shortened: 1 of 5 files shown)


packages/falcon-front-kit/src/Checkout/CheckoutProvider.tsx
```diff
@@ -1,17 +1,36 @@
...
 export const CheckoutProvider: React.SFC<CheckoutProviderProps> = props => {
-  const { children, initialValues, billingSameAsShipping } = props;
+  const { children, initialValues = {}, billingSameAsShipping, stepsOrder, autoStepForward } = props;
 
-  const [step, setStep] = useState<keyof typeof CheckoutStep>('Email');
+  /**
+   * Returns next possible step for `step` based on available `CheckoutFlow`, or `undefined` when no more steps
+   * @param step
+   */
+  const getNextStep = (step: CheckoutStepType): CheckoutStepType | undefined => {
```
Borales: @tutok just a question - could we keep both `getNextStep` and `getNextStepForValues` outside of this Provider? it was nice to have such helper methods for SearchContext and being able to use them without a complete Context setup.

tutok: even that you need to have some "basic setup". I mean, you need to have available steps in order (`stepOrder`).

PR #755, opened 2019-12-27
Dropping Node 8 support

Borales:
This PR raises the "required minimal node version" check from 8.10 to 10.15 (because `8.x`reaches the end-of-life within a couple of days).

The reason for `10.15` - it was released about a year ago, so there's a higher chance that developers would have it installed on their local machines, hence we would give them less stress during the upgrade.


nimobeeren:
Seems like something is wrong with the CLA plugin? @Borales

PR #764, opened 2020-01-13
Improving Coupon type across the schema

Borales:

Feature. Having a `couponCode` string field on Cart/Order is not enough for advanced templates.
This PR introduces `Coupon` type for Shop Extension.

This PR allows using multiple coupon codes for the cart (if backend supports it, like WooCommerce). Allows removing a specified coupon using its code.

Yes

### DIFF
--- a/packages/falcon-magento2-api/src/index.ts
+++ b/packages/falcon-magento2-api/src/index.ts
@@ -879,7 +879,7 @@ module.exports = class Magento2Api extends Magento2ApiBase {
         {
           code: totalsData.couponCode,
           name: totalsData.couponCode,
-          discount: totalsData.discountAmount
+          discount: totalsData.totalSegments.find(({ code }) => code === 'discount').value
         }
       ];
     }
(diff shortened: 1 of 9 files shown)


packages/falcon-magento2-api/src/index.ts
```diff
@@ -1111,10 +1120,20 @@ module.exports = class Magento2Api extends Magento2ApiBase {
   reduceOrder(response) {
     const order = this.convertKeys(response);
     const { extensionAttributes = {}, payment = {}, entityId, incrementId, items, ...orderRest } = order;
+    const coupons = [];
+
+    if (order.discountAmount) {
```
tutok: I am not really sure if we can rely on `discountAmount` in order to add a coupon code.
Because you can get a discount if you would like to buy some specific amount of items, (eg. 10 pieces of t-shirt)

Borales: @tutok fixed by extracting this code from `totalSegments` data

Borales: for order - we only get `discountAmount` and `discountDescription` (as I understand - only 1 cart rule can be applied on Magento at a time, so it'll be our "coupon code discount")

PR #766, opened 2020-01-15
FIx a11y issues with the number input

### LINKED ISSUE #735: Keyboard support for NumberInput
**Is your feature request related to a problem? Please describe.**
I would like to use the `NumberInput` with the keyboard. I can place the cursor in the text field, but not adjust it by typing.
image

**Describe the solution you'd like**
- Allow typing in a number by selecting the text field.
- Update state after the input loses focus.
- Use up/down (left/right?) arrows to adjust the number one by one (or larger increments?).

hobadams:
I've removed the read-only as a default prop (it was added here - https://github.com/deity-io/falcon/commit/43a72f0a9248ccfaab7adae5d682aa94c3aad20b#diff-3c6e5855bfb17f965488c8202f5af42f).

I've also removed the `aria-hidden` from the qty buttons. In my opinion, they should be able to be picked up by assistive technology...especially if the input is read-only.

I've also added aria labels and disable the input buttons when they have reached their min / max value

Fixes issue:
Closes #735 

**Other information:**

### DIFF
--- a/packages/falcon-ui/src/components/NumberInput.tsx
+++ b/packages/falcon-ui/src/components/NumberInput.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { I18n } from '@deity/falcon-i18n';
 import { themed, extractThemableProps } from '../theme';
 import { Box } from './Box';
 import { Icon } from './Icon';
@@ -21,6 +20,8 @@ function triggerChange(element: any, value: any) {
 
 type NumberInputInnerDOMProps = {
   invalid?: boolean;
+  increaseText?: string;
+  decreaseText?: string;
 } & React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>;
 
 class NumberInputInnerDOM extends React.Component<NumberInputInnerDOMProps> {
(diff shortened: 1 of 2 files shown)


packages/falcon-ui/src/components/NumberInput.tsx
```diff
@@ -84,13 +84,13 @@ class NumberInputInnerDOM extends React.Component<NumberInputInnerDOMProps> {
 
     return (
       <Box {...themableProps} className={className}>
-        <button type="button" aria-hidden onClick={this.stepDown} className="-inner-input-step-down-element">
+        <button type="button" onClick={this.stepDown} aria-label="Decrease quantity" className="-inner-input-step-down-element">
```
hobadams: I know this text should be run through a translation function.....see the PR description.

hobadams:
Closing while i fix the translations

hobadams:
I need to look into why the deps aren't installing correctly

hobadams:
> I'm afraid we cannot merge this pull request. Falcon UI is supposed to be agnostic in each possible aspect while adding falcon-i18n breaks this rule as it forces users to use our translation system. What's more, it also adds all the dependencies of falcon-i18n which will increase the bundle size.
> 
> We can add required attributes to solve the a11y issues, but the texts would have to be passed as already translated strings.

@mateuszwozniak - I've replaced this in favor of props. I've added default props in English rather than making them required. Are you ok with this?

packages/falcon-ui/src/components/NumberInput.tsx
```diff
@@ -79,18 +81,31 @@ class NumberInputInnerDOM extends React.Component<NumberInputInnerDOMProps> {
...
+          onClick={this.stepDown}
+          aria-label={decreaseText}
+          className="-inner-input-step-down-element"
+          disabled={currentValue <= min}
+        >
           <Icon src="numberInputDown" fallback="−" />
         </button>
 
-        <input ref={this.inputRef} min={5} type="number" {...rest} />
+        <input ref={this.inputRef} min={min} value={value} type="number" {...rest} />
```
nimobeeren:
```suggestion
        <input ref={this.inputRef} min={min} max={max} value={value} type="number" {...rest} />
```
I think `max` wasn't properly set because it was destructured from props.

PR #788, opened 2020-11-16
Update scopes

mateuszwozniak:
Change scopes from `@deity` to` @deity-io`


Borales (approved):

hobadams (approved):
I've just skimmed through to check nothing dodgy is committed. I've not tested it

